Issue #12, opened 2021-05-25
Enhancement: Edit Client Name


drtech981:
There's no way to edit a client's name once you hit submit. Granted I can just delete and re-add the client, it would be easier if there was an option to edit the client name !

mikekosulin:
Until then you can do that by changing it in the `wg0.json` and `wg0.conf` and restarting wg-easy container

rapejim:
> Until then you can do that by changing it in the `wg0.json` and `wg0.conf` and restarting wg-easy container

If only want edit name, I edit only `wg0.json` and I restart wg-easy container.

rapejim:
@drtech981 can you try the changes for edit name and/or last ip number of my pull request?

https://github.com/WeeJeWel/wg-easy/pull/38

You can try this changes using temporally the docker image of my fork `rapejim/wg-easy`.
(IMPORTANT: this imagen is only for testing this changes, probably I delete it if WeeJeWel finally approve and merge the PR 😅).

drtech981:
@rapejim 
I tested it and it works absolutely well ! Waiting for the PR to be merged.

drtech981:
@rapejim can you also add an option for multiple user profiles so that each person can login and change only his or her devices ?

rapejim:
> @rapejim can you also add an option for multiple user profiles so that each person can login and change only his or her devices ?

This is a "mayor" update (imho), and maybe you must requested with a new issue, 😅

drtech981:
@rapejim https://github.com/WeeJeWel/wg-easy/issues/10 I had opened this request but had it closed. 
Can you reopen the issue and work on it ?

Greatly counting towards your contribution !

WeeJeWel:
I wonder what you want non-admin users to be able to do?

drtech981:
Admin can choose to add local users (authelia/sso) / LDAP auth, and each user can add or remove their devices.

I use WG extensively to access my clinic's PBX,  and so do my colleagues. However a single page lists all devices, and many a times there's an accidental deletion or someone disables someone's else device. Would be great if one person is responsible for their set of devices only.

I know, a very niche case, but work on it would be great.

rapejim:
> Admin can choose to add local users (authelia/sso) / LDAP auth, and each user can add or remove their devices.
> 
[...]

PR #38, opened 2021-06-19
Add last ip number selection

rapejim:
- Added last IP number selection for new client (auto if empty)
- Added sort clients on webUI by last IP number

image
image
image


aniqueta:
I like this change. I was migrating over an old config, and so just went through to manually set things in the JSON… It would also be nice to be able to set the "server" IP (which I did by editing WireGuard.js).

WeeJeWel:
This is very nice functionality. I've given it some thought and came up with an (imho) better UX:

* [New Client] dialog doesn't change
* When hovering over the IP address in the list, a small edit-icon (pencil) appears. Clicking it makes the IP address an inline-textfield.
* We do the same for name and solve #12

WeeJeWel:
FYI I won't merge this until it works like I suggested, because this code is too different :)

rapejim:
> FYI I won't merge this until it works like I suggested, because this code is too different :)

I don't have enough knowledge to develop your suggestion "BLABLABLABLA" but I just uploaded a commit with the changes to edit the client using a pencil icon next to the delete icon and make the changes in a modal window like the new client.

image
image
image

I hope that these changes I have made are enough to your satisfaction to merge them.


Ah! si quieres probar los cambios fácilmente, en mi fork yo he unido los cambios en la rama de producción y por tanto puedes probarlos tan fácilmente como usando (aunque sea temporalmente) my imagen de docker `rapejim/wg-easy`

WeeJeWel:
This is now included in release 2.

PR #39, opened 2021-06-21
Graceful iptables changes

aniqueta:
I think this is a more standard and graceful way of doing the iptables changes. It uses WireGuard's PostUp and PostDown functions. When the tunnel goes down, the up-tables changes are reversed. This is useful for anyone who installs this without Docker.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -52,10 +52,6 @@ module.exports = class WireGuard {
 
         await this.__saveConfig(config);
         await Util.exec('wg-quick up wg0');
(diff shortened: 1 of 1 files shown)


WeeJeWel:
Thanks, but the scope of this project is not to support users that don't use Docker.

aniqueta:
@WeeJeWel Understand keeping the scope focused and clean. 

Just so you know, it was super easy to adapt this to work well without Docker. The above change and some manual build and system service configuration. Some systems do not perform well with Docker (VM on macOS, networking stack, etc.), and it's also nice to just layer this on as a frontend to an existing configuration. 

I do think that regardless of Docker, the change I suggested is the more 'correct' or 'standard' way to make iptables changes with WireGuard.

Anyhow, understand the choice here, and thanks for considering.

PR #59, opened 2021-08-30
Add a stable Docker tag

FuzzyMistborn:
Fixes https://github.com/WeeJeWel/wg-easy/issues/56

Adds a "stable" tag that's automatically built every week instead of daily.


FuzzyMistborn:
Also, as another suggestion you could start making releases and triggering the stable build off that, like I have here: https://github.com/FuzzyMistborn/wg-easy/blob/main/.github/workflows/deploy-stable.yml

PR #74, opened 2021-09-22
Update readme.md add volume for persistent data

WillemWever:
Hi, would you like to update the readme with volume information. I think that users should mount /etc/wireguard to a local folder to persistent changes in the config.


gaby:
@WillemWever Can you close this an open an Issue instead?

PR #81, opened 2021-10-07
Theming (black theme), NAT (as per pull 43 from @principis), details switch for black theme, WG_PORT reflected in WireGuard.js

BankaiNoJutsu:
Hey!

I really like what you did here, so i made some changes, trying to improve a bit.
Feel free to keep what you want/like.

- Theming (black theme added)
- NAT (as per pull 43 from @principis thanks!)
- Details switch for both themes (shows more infos, like endpoint and allowed ip's list in peer config)
- WG_PORT reflected in WireGuard.js, 51820 was hardcoded?
- New logo for black theme, as contrast was bad with the svg

Cheers!

2021-10-07 22_57_10-Window


gaby:
@WeeJeWel Thoughts on this?

WeeJeWel:
It's nice :)

However the theme should be CSS-only, not a different HTML file.

Also the WG_PORT shouldn't need to be changed.

I prefer one PR per feature and not bundles :)

andrey0001:
IPV6 is very important!

PR #83, opened 2021-10-11
Normalize config file name for download to make it work on Windows

vojtad:
Client name is derived from config file name when importing configuration to WireGuard on Windows and it only accepts names satisfying `/^[a-zA-Z0-9_=+.-]{1,32}$/`. For more information look at https://github.com/WireGuard/wireguard-windows/blob/af60ab229954519b8295bb3ef453231f4d3b9087/conf/name.go#L24.

This PR implements a name normalization where invalid characters are replaced by `-`.

Without this PR WireGuard on Windows throws `Tunnel name ‘%s’ is invalid.` error when importing config file with invalid characters in its file name.


WeeJeWel:
This is absolutely a great addition, thank you!

PR #90, opened 2021-10-26
Allow custom rules using `FIREWALL_RULES`

scottdotau:
Relates to https://github.com/WeeJeWel/wg-easy/pull/43, allows the user to pass through `FIREWALL_RULES` environment variable, which can be set to `false` to disable completely.

Note that I'm not really sure that passing through as an environment var is the best method, but just wanted to get a POC that works up to explains what I mean.

Did some testing and all works well, but note: 
- This does no checks on if the 'rules' are actually valid, it just passes through each line to the Util.exec command as-is, meaning anything can be run.
- If a command is invalid, the container will break on startup (which is probably not a terrible thing anyway).

For my use case specifically, I needed my connection to MASQUERADE through tun0 rather than eth0, therefore:

`FIREWALL_RULES=iptables -t nat -A POSTROUTING -s 10.11.11.0/24 -o tun0 -j MASQUERADE;iptables -A INPUT -p udp -m udp --dport 51820 -j ACCEPT; iptables -A FORWARD -i wg0 -j ACCEPT; iptables -A FORWARD -o wg0 -j ACCEPT`

Works for me

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -13,4 +13,13 @@ module.exports.WG_DEFAULT_ADDRESS = process.env.WG_DEFAULT_ADDRESS || '10.8.0.x'
 module.exports.WG_DEFAULT_DNS = typeof process.env.WG_DEFAULT_DNS === 'string'
   ? process.env.WG_DEFAULT_DNS
   : '1.1.1.1';
+module.exports.FIREWALL_RULES = process.env.FIREWALL_RULES === 'false'
+  ? false
+  : process.env.FIREWALL_RULES?.split(';') || [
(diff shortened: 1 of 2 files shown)


niem94:
There is no need for this. You already have the WG_POST_UP environment variable, you can use to override the iptables config used by default. I just did this for the purpose of changing the default interface to eth1, like you want to change yours to tun0.

```
WG_POST_UP=iptables -A FORWARD -i %i -j ACCEPT; iptables -A FORWARD -o %i -j ACCEPT; iptables -t nat -A POSTROUTING -o tun0 -j MASQUERADE
```

Should do the trick :-)

scottdotau:
Yeah, these must have been added recently!

PR #98, opened 2021-10-28
Update README.md

Neptunespace:
The most simple way to install the container for people start with docker ;)


WeeJeWel:
Not really, it doesn't auto-create the config directory :)

WeeJeWel:
Updated :)

WeeJeWel:
> Not really, it doesn't auto-create the config directory :)

I guess I was wrong, it does create the directory!

PR #99, opened 2021-10-31
Removed sensitive info (private keys) from being logged

joshuakraitberg:
Private keys should not be visible in logs.

Example of new output:
```
$ wg genkey
$ echo *HIDDEN* | wg pubkey
$ wg genpsk
```


WeeJeWel:
Now fixed.

PR #125, opened 2021-11-19
Use relative paths for client

m1k1o:
I wanted to deploy this GUI on `https://domain.tld/wg-easy/`, but it was not working properly, since some links were absolute and expecting `/api` and `/js` on root of my domain.

I changed them to relative paths, so tht such deployments are possible. Or would you prefer having it environment vairable e.g. `BASE_PATH=/wg-easy/` ?


WeeJeWel:
I prefer not, makes it harder to maintain and add routing later.

m1k1o:
In that case, all resources should be converted to / root path. Some of them are missing it (e.g. css). I think it should be at least unified.

PR #131, opened 2021-11-23
README : Improve Docker Installation instructions

nicolasverlhiac:


WeeJeWel:
Why is this better?

nicolasverlhiac:
Hi @WeeJeWel,

The convenience script is not recommended for production usage.
https://docs.docker.com/engine/install/

I thought it would help to have both approaches available for the Docker setup.

What do you think about it?

Thank you

ffoxin:
> Hi @WeeJeWel,
> 
> The convenience script is not recommended for production usage. https://docs.docker.com/engine/install/
> 
> I thought it would help to have both approaches available for the Docker setup.
> 
> What do you think about it?
> 
> Thank you

Are you sure that adding static manual to README.md is better than adding just a link to the up-to-date manual?

nicolasverlhiac:
It is indeed possible to keep only this part: 

---
### 1. Install Docker

If you haven't installed Docker yet, install it by running:
You can follow this instruction, but don't forget to take a look to the [Docker documentation](https://docs.docker.com/engine/install/).

PR #145, opened 2021-12-15
Add iptables support for WG_ALLOWED_IPS

Retloldin:
By default this WireGuard server implementation doesn't filter the destination IP addresses that the client requests.

Only with the AllowedIPs field of the configuration file the end users can access any destination if they edit it, creating a false security in the container configuration.

Adding the use of iptables as a filter for accessible networks brings up the security that the WG ALLOWED_IPS field to the user.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -53,12 +53,27 @@ module.exports = class WireGuard {
           debug('Configuration generated.');
         }
 
+        if (WG_ALLOWED_IPS) {
+          var ALLOWED_IPS = WG_ALLOWED_IPS.split(',');
+        }else{
+          var ALLOWED_IPS = new Array();
+        }
+
         await this.__saveConfig(config);
         await Util.exec('wg-quick down wg0').catch(() => { });
         await Util.exec('wg-quick up wg0');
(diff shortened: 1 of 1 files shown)


psycmos:
Subscribed ;)

psycmos (changes requested):
It's possible to include Variable Port Restrict like Allowed IP?

Retloldin:
> It's possible to include Variable Port Restrict like Allowed IP?

This would require changing the input format of the WG_ALLOWED_IPS variable or creating a new one to define the ports to be filtered with iptables.

The proposed change in this request doesn't require any changes from the end users in case of a Docker update.

psycmos:
> > It's possible to include Variable Port Restrict like Allowed IP?
> 
> This would require changing the input format of the WG_ALLOWED_IPS variable or creating a new one to define the ports to be filtered with iptables.
> 
> The proposed change in this request doesn't require any changes from the end users in case of a Docker update.

You are right, it require a new variable, i will create a new request :D
Maybe the creators can change the default template ;)

WeeJeWel:
You can now overwrite all iptables with env var `WG_POST_UP`.

PR #150, opened 2021-12-22
Fix Relative links and allow use of WG_PATH variable

jdeath:
This PR fixes the relative links (same as #125) and also allows use of the WG_PATH variable. The system will properly load and use config files not saved in /etc/wireguard/

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -43,4 +43,4 @@ jobs:
       with:
         push: true
         platforms: linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64/v8
-        tags: weejewel/wg-easy:latest, weejewel/wg-easy:${{ env.RELEASE }}
+        tags: jdeath/wg-easy:latest, jdeath/wg-easy:${{ env.RELEASE }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,8 +10,6 @@ on:
       - staging/**
       - production
       - production/**
-  schedule:
-    - cron: "0 12 * * *"
 
 jobs:
   deploy:
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -106,7 +106,7 @@ AllowedIPs = ${client.address}/32`;
 
   async __syncConfig() {
     debug('Syncing config...');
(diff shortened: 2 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -43,4 +43,4 @@ jobs:
       with:
         push: true
         platforms: linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64/v8
-        tags: weejewel/wg-easy:latest, weejewel/wg-easy:${{ env.RELEASE }}
+        tags: jdeath/wg-easy:latest, jdeath/wg-easy:${{ env.RELEASE }}
```
WeeJeWel, line 46: ?

src/lib/WireGuard.js
```diff
@@ -54,8 +54,8 @@ module.exports = class WireGuard {
         }
 
         await this.__saveConfig(config);
-        await Util.exec('wg-quick down wg0').catch(() => { });
-        await Util.exec('wg-quick up wg0');
+        await Util.exec(`wg-quick down ${WG_PATH}/wg0.conf`).catch(() => { });
+        await Util.exec(`wg-quick up ${WG_PATH}/wg0.conf`);
```
WeeJeWel, line 58: Why is this useful? We're using Docker so who cares about the path :)

src/www/index.html
```diff
@@ -3,7 +3,7 @@
 
 <head>
   <title>WireGuard</title>
-  <link href="/css/vendor/tailwind.min.css" rel="stylesheet">
+  <link href="css/vendor/tailwind.min.css" rel="stylesheet">
```
WeeJeWel, line 6: I prefer `./` to indicate it's a relative path.

WeeJeWel (changes requested):

.github/workflows/deploy.yml
jdeath: This was so I could use and test it. I didn't know how to remove this file...

src/lib/WireGuard.js
jdeath: I am using the container inside homeassistant (stripped down docker). I cannot mount another directory to /etc/wireguard/.

src/www/index.html
jdeath: I can make the change. It worked, so I did it.

WeeJeWel:
Relative links are now added.

Honestly I have no interest in WG_PATH, because it's mounted using Docker anyway.

PR #154, opened 2021-12-26
add optional client profile MTU var

DerDanilo:
Add feature to configure optional client profile MTU.
https://github.com/WeeJeWel/wg-easy/issues/95

I am not sure if the if block is to your liking since javascript I am not that familiar with javascript.
Please check and update the code if desired. Thanks!


DerDanilo:
The check failed at the app.js which is not a file I changed.  What is the cause for this?

Additionally I am not sure if the way of checking for WG_MTU is recommended. It is only supposed to be written into the generated profiles when configured. 

### Currently
```
${WG_DEFAULT_DNS ? `DNS = ${WG_DEFAULT_DNS}` : ''}
if (typeof ${WG_MTU} !== 'undefined' || ${WG_MTU} !== null) {
MTU = ${WG_MTU}
}
```

### Alternativ?

Not sure if `?` catched `null` also?
```
${WG_DEFAULT_DNS ? `DNS = ${WG_DEFAULT_DNS}` : ''}
${WG_MTU ? `MTU = ${WG_MTU}` : ''}
```

WeeJeWel (approved):

PR #160, opened 2022-01-08
Improved some styles and fixed some issues with the modals in mobile.

ApexioDaCoder:
This PR addresses some issues with the modals on mobile, switches to the tailwind CDN script, and adds animated toggles instead of just the switching ones.


WeeJeWel:
Please don't use the CDN, that makes the front-end internet-dependent!

Issue #168, opened 2022-01-12
Support for docker secrets on the PASSWORD variable and the WG_HOST variable.


niem94:
Currently, docker secrets are not working when used with the above-mentioned environment variables. I would like these values to stay secure, why it would be nice if docker secrets were supported.

I will look into making a PR shortly 👍🏻

PR #169, opened 2022-01-12
Added docker secrets support for PASSWORD and WG_HOST in Docker Swarm

niem94:
#168 

@WeeJeWel Can you test if this works as expected for normal use cases? I added a docker-compose.dev-secrets.yml file as well with some guidance on how to test it with docker swarm. 👍🏻

### DIFF
--- a/docker-compose.dev-secrets.yml
+++ b/docker-compose.dev-secrets.yml
@@ -18,7 +18,7 @@ secrets:
     external: true
 
 # require docker swarm to test
-# docker swarm int
+# docker swarm init
 # printf "password" | docker secret create wireguard_password -
 # printf "domain.com" | docker secret create wireguard_host -
 # docker stack deploy -c docker-compose.dev-secrets.yml wg-easy
--- a/docker-compose.dev-secrets.yml
+++ b/docker-compose.dev-secrets.yml
@@ -1,0 +1,24 @@
+version: "3.8"
+services:
+  wg-easy:
+    image: wgeasy
+    command: npm run serve
+    volumes:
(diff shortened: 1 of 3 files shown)


WeeJeWel:
Hm. This adds complexity for a specific use case, and I don't really understand how this is safer?

WeeJeWel:
You say ‘the repo’, as in, do you commit your docker compose file?

In that case, just put them in ‘.env’!

niem94:
Yes I do, and I see setting the env variables elsewhere could be one way to make it work, but I would still prefer if I would be able to set the env variables from docker secrets. This is what I am used to do with all my stacks, and I would like to be consistent. There are several reasons why:

1. Docker will tell me if I have not set the required secrets for a stack to run.
2. It is a solution that solely requires docker. Low maintenance
3. It provides a layer of abstraction between the containers and their environment.

Is there no interest in supporting this? Otherwise I do not mind making a fork and maintaining a version of wg-easy that is more centered around my usecase.

niem94:
I just researched this a bit, as I was a bit confused if I was missing something. It seems I was. I was not aware that one could use env variables like so: <https://docs.docker.com/compose/environment-variables/>

I will definitely give it a look, and see if it is what I am looking for 👍🏻 

Closing this issue for now.

carmatana:
Hello,

This is not working

declaring a variable in .env file is not safe, in this link is he explanantion

https://docs.docker.com/engine/swarm/secrets/

Use of secrets is the solution

https://docs.docker.com/compose/use-secrets/

Would be nice to have this added level of security

PR #197, opened 2022-01-31
Docker compose example

tamet83:
Hi, I inserted a Docker compose example.
Hope you agree it will be useful for many others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,41 @@ $ docker run -d \
 
 The Web UI will now be available on `http://0.0.0.0:51821`.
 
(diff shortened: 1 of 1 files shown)


ffoxin:
There is [docker-compose](https://github.com/WeeJeWel/wg-easy/blob/master/docker-compose.yml) example in the root directory, it could be easily downloaded as is. Why do we need one more example in the readme file that is also required copy-paste?

tamet83:
Sorry, I missed that information. Thanks for your time :)

PR #216, opened 2022-03-05
feat(WireGuard): return client data by createClient.

murka:


WeeJeWel:
Why is this useful? :)

divadsn:
> Why is this useful? :)

For creating integrations/scripts using the API, before you had to get all clients after creating to find the client ID 😉

PR #224, opened 2022-03-15
Update remplate for easy integarion with prometheus_wireguard_exporter

mfominov:
There is a good project https://github.com/MindFlavor/prometheus_wireguard_exporter

This update is just for compatibility.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -105,6 +105,8 @@ PostDown = ${WG_POST_DOWN}
 
 # Client: ${client.name} (${clientId})
 [Peer]
+# friendly_name = ${client.name}
+# (${clientId})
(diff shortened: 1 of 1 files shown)


benisai:
I would love this @WeeJeWel

WeeJeWel:
I'd prefer not to have compatibility with 3rd party tools unless the WireGuard docs mention this as 'approved' or so.

mfominov:
@WeeJeWel u using same mechanism to get stats in web ui https://github.com/WeeJeWel/wg-easy/blob/master/src/lib/WireGuard.js#L151-L183
and the exporter https://github.com/MindFlavor/prometheus_wireguard_exporter#systemd-service-file
`
wg show wg0 dump
`
is the only way to get statistics about connection

PR #241, opened 2022-04-06
Button to toggle charts visibility

madiskar:
Greetings!

First off, thanks for making this project. It has greatly simplified my WireGuard deployments!

However, in one my deployments, I currently have 12 devices enrolled (and increasing). This has presented a performance issue with the live Rx/Tx graphs. On my M1 Macbook Pro (and other devices as well), using Brave or Firefox, in times where there is a lot of network activity, the **UI frequently stays unresponsive**. Low network (and therefore chart) activity seems to improve things but it's still far from ideal.

I believe the same issue is described here as well: https://github.com/WeeJeWel/wg-easy/issues/187

My solution is to add a "Charts" button at the top right which simply toggles the visibility of live charts. 
charts

Testing on my Macbook, hiding the charts seems to _drastically_ improve UI responsiveness. A better solution would probably be to look into the charting library or switch to a different one, but unfortunately I don't have the time to dig into this at the moment.

Visibility state is also stored in localStorage for convenience.

Icon source: https://www.svgrepo.com/svg/17087/graph

Let me know what you think,
~ MK

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -63,6 +63,19 @@
                 </svg>
                 <span class="text-sm">New</span>
               </button>
+
+              <button @click="toggleCharts();"
+                class="hover:bg-red-800 hover:border-red-800 hover:text-white text-gray-700 border-2 border-gray-100 py-2 px-4 rounded inline-flex items-center transition">
+                <svg class="w-4 mr-2" inline xmlns="http://www.w3.org/2000/svg" fill="currentColor" viewBox="0 0 455 455">
+                  <path d="M415,102.509c-22.091,0-40,17.909-40,40c0,5.542,1.128,10.821,3.166,15.62l-83.791,83.792
+                    c-4.799-2.038-10.078-3.167-15.621-3.167s-10.822,1.129-15.621,3.167l-50.053-50.053c2.038-4.799,3.166-10.078,3.166-15.621
+                    c0-22.091-17.909-40-40-40c-22.091,0-40,17.909-40,40c0,5.542,1.128,10.821,3.166,15.62l-83.792,83.791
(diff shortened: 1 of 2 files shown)


karlerkijy:
Would love to see this merged. :heart:

WeeJeWel:
I'd rather fix the chart's performance :)

karlerkijy:
It would be really nice to turn off the chart because for our use case it is too much noise.

madiskar:
I agree, that chart performance is something that should be looked at. However as @karlerkijy mentioned, UI noise is also a factor here -- while I love who the charts looks visually, I think having an option to turn them off would still be useful for folks who prefer a more cleaner look :)

WeeJeWel:
Maybe only show one on hover?

madiskar:
Yeah, that's also an option, it should solve both issues.

Although I'm not sure how big of a concern mobile UX is for this project, hover would not be best option for mobile I think (I would _assume_ most people use this UI with a pointing device).

Another alternative that's perhaps a bit more mobile friendly, would be to add another icon button inside each client section, that would toggle the chart for that client. Though now that I look at the current mobile layout, it could definitely use some improvement as well :D But that's a bit out of scope for this PR.

Personally, I'm good with implementing either version, or keeping everything as is. Let me know which one you prefer.

EugeneVl:
Would love to see this merged too) charts perfomance with high number of clients make chrome very laggy)

tsugliani:
Would love to see either this option, or better charts performance for sure.

I'll try this PR for sure, for me the main "live" information I'm looking for is to see if someone is connected or not, and it's last connection state if not, but the charts do add a very nice feel to it.

Same issue here with 50+ devices, it's quite slow ;-)

WeeJeWel:
Performance issues have been fixed.

mikekosulin:
image

better, but the same experience on 6-nightly and latest.

PR #257, opened 2022-05-08
Update WireGuard.js ( WG_PORT variable fix ) 

cuzrawr:
Fix for WG_PORT variable, because it was hardcoded to 51820 before.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -93,7 +93,7 @@ module.exports = class WireGuard {
 [Interface]
 PrivateKey = ${config.server.privateKey}
 Address = ${config.server.address}/24
-ListenPort = 51820
+ListenPort = ${WG_PORT}
 PostUp = ${WG_POST_UP}
 PostDown = ${WG_POST_DOWN}
 `;


WeeJeWel:
The port is mapped by Docker.

cuzrawr:
> The port is mapped by Docker.

Hi. I can’t understand why this port should be hard-coded, because in this way we lose flexibility. Please look inside this config file, ( https://github.com/WeeJeWel/wg-easy/blob/master/src/config.js )  we already have here fallback 51820 WG_PORT variable.
So its always fallback due to "OR" statement. Like other config params.

this variable ( WG_PORT ) is also specified   [here](https://github.com/WeeJeWel/wg-easy#options)  in your project description file, on main page, its confusing, because inside docker-compose if you set this variable its do nothing. So my pull fixing this behavior.

And on other side i wanna say port forwarding from host to docker is always have significant overhead to critical network services like a warguard. Because its working through NAT. But services like this requires speed.
, network_mode: "host"          - better to use instead.
This option is faster.

Also I can share my production  docker-compose.yml  if you wanna look for example.

PR #178, opened 2022-01-20
[Feature] Add dark mode 🌙

Teraskull:
This PR contains the following changes:

* Add dark mode using the new `dark:` class in Tailwind 3.
* Add theme switch button.
* Replace all dark red colors with pastel alternatives in dark mode.
* Fix static HTML elements flashing when reloading page.
* Fix mobile layout and overflowing elements.
* Update `tailwind.min.css` to only include used classes, which reduces the size from `2859KB` to just `20KB`.
> Future changes to the style will require to use a CDN for development and tree shaking the file again.

This PR also changes a few things in light mode:
* The toggles are now green (Red is not a good option for `On` state).
* Removed the double border from the `Clients` header.
* Removed the bottom border/divider from the last table child.

## Screenshots:
wireguard_dashboard
wireguard_login
wireguard_modal

### Before vs After (light/dark):
image

All elements should be properly styled in dark mode, including the Update banner. If I missed anything, let me know. (Closes #163)

PR #260, opened 2022-05-18
Fix modals flashing on reload

### LINKED ISSUE #251: Splash screen before Web UI logon prompt
Every time I opened the Web UI, it is showing me this initially for just a few milliseconds and then prompting to the logon screen, is this normal?
image

Teraskull:
Fixes #251.

This PR is separated from #178.

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -107,7 +107,7 @@
                 <div class="flex-grow">
 
                   <!-- Name -->
-                  <div class="text-gray-700 group" :title="'Created at ' + dateTime(new Date(client.createdAt))">
+                  <div class="text-gray-700 group" :title="'Created on ' + dateTime(new Date(client.createdAt))">
 
                     <!-- Show -->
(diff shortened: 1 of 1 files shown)


src/www/index.html
```diff
@@ -101,7 +107,7 @@ <h2 class="text-sm text-gray-400 mb-10"></h2>
                 <div class="flex-grow">
 
                   <!-- Name -->
-                  <div class="text-gray-700 group" :title="'Created at ' + dateTime(new Date(client.createdAt))">
+                  <div class="text-gray-700 group" :title="'Created on ' + dateTime(new Date(client.createdAt))">
```
WeeJeWel, line 110: Why has this changed? `at` is used for exact times, `on` for days/months/years.

WeeJeWel (changes requested):

src/www/index.html
Teraskull: That's why it's changed:

image

I'm not sure if it is displayed in a different format for you, in which case you would want to handle the format order.

PR #285, opened 2022-06-14
Update config.js

L719dhh:
POST_UP Postup did not use the variable WG_PORT


WeeJeWel:
WG_PORT is only used for the clients information. Docker maps this port.

PR #312, opened 2022-07-08
Added option for Interface choice for the webserver to listen on

maxgestic:
Added the option to let the user specify an interface for the webserver to run on since by default it will only run on eth0, this lets you set add -e INTERFACE=wlan0 for example to the docker run script and makes it work with any interfaces. It was mentioned in this issue https://github.com/WeeJeWel/wg-easy/issues/309


maxgestic:
Nevermind im dumb lol

PR #323, opened 2022-07-22
CoreDNS integration - use docker host's domain names resolving capabilities

petr-ujezdsky:
Hi,

I have seen this functionality in https://github.com/linuxserver/docker-wireguard#parameters project and found it super useful. **It forwards domain names resolving to the docker host.**

This comes really handy if you run your own DNS server (eg. `pi-hole`) and you do not want to hard-code it's IP into container config. Or your server itself has some complex domain names resolving scheme and you want clients to use it.

I would set it as default (as is in the source project) but that would be breaking change (your default is `1.1.1.1`).

The little downside is added size of 46MB.
```
$ docker images | grep wg
wg-easy                               latest         e35491bbd4a7   5 minutes ago    188MB
weejewel/wg-easy                      latest         c90b263d8a38   3 weeks ago      142MB
```

The CoreDNS installation always uses the latest version. This could be changed for more stable releases of `wg-easy`.
CoreDNS is also available in docker image format. This could ease the install using multistage build. However only `x86` architecture is dockerized, so the installation for `arm` architecture would have to be done downloading it from internet using `curl`. The source project [use](https://github.com/linuxserver/docker-wireguard/blob/master/Dockerfile#L51) `curl` download for both platforms making it much more easier to manage. I have reused the same code.

It is also possible to export the port 53 and allow anyone on the network to use the server's domain names resolving capabilities.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,23 @@ RUN npm i -g nodemon
 # Install Linux packages
 RUN apk add -U --no-cache \
   wireguard-tools \
-  dumb-init
+  dumb-init \
+  curl
+
+# Install latest coredns
+RUN mkdir /coredns && \
+  COREDNS_VERSION=$(curl -sX GET "https://api.github.com/repos/coredns/coredns/releases/latest" \
+	| awk '/tag_name/{print $4;exit}' FS='[""]' | awk '{print substr($1,2); }') && \
+  curl -o \
+	  /tmp/coredns.tar.gz -L \
+	  "https://github.com/coredns/coredns/releases/download/v${COREDNS_VERSION}/coredns_${COREDNS_VERSION}_linux_amd64.tgz" && \
+  tar xf \
+	  /tmp/coredns.tar.gz -C \
+	  /coredns && \
+  rm -rf /tmp/coredns.tar.gz
+
+COPY coredns/Corefile /coredns
+COPY entrypoint.sh /
 
 # Expose Ports
 EXPOSE 51820/udp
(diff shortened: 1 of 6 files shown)


archont94:
I would rather put `dumbinit` in ENTRYPINT and bash script in CMD, as suggested in [dumb-init how-to](https://github.com/Yelp/dumb-init):

```
ENTRYPOINT ["/usr/bin/dumb-init", "--"]
CMD ["bash", "/entrypoint.sh"]
```
And don't use `dumb-init` inside `entrypoint.sh` anymore.

petr-ujezdsky:
@WeeJeWel do you use multiarch builds? If so, do you know how to use different URL per architecture to download the CoreDNS binary from web?

Eg.
* amd64: https://github.com/coredns/coredns/releases/download/v${COREDNS_VERSION}/coredns_${COREDNS_VERSION}_linux_amd64.tgz
* arm64: https://github.com/coredns/coredns/releases/download/v${COREDNS_VERSION}/coredns_${COREDNS_VERSION}_linux_arm64.tgz

archont94:
It seems like it's beeing builded for multi-archs: [linux/amd64, linux/arm/v6, linux/arm/v7 and linux/arm64](https://hub.docker.com/r/weejewel/wg-easy/tags)

gaby:
@petr-ujezdsky There's no reason to build CoreDNS. It's possible to download the proper binary during wg-easy build. 

For example, this is how to download/extract Helm:
```
ARG HELM_VERSION
ARG TARGETOS
ARG TARGETARCH

wget -q https://get.helm.sh/helm-v${HELM_VERSION}-${TARGETOS}-${TARGETARCH}.tar.gz -O - | tar -xzO ${TARGETOS}-${TARGETARCH}/helm > /usr/local/bin/helm
```

The TARGETOS args get passed to the build if doing Multi-Arch builds with `docker buildx`.

gaby:
Also Alpine alreasy publishes a CoreDNS package

```
FROM alpine:3

RUN apk update && apk add coredns
```

Downloading the binary as suggested in my previous post is better since that will get the latest release.

PR #328, opened 2022-07-26
wireguard-go support

### LINKED ISSUE #327: wireguard-go support for VPSes without wireguard kernel module enabled.
Cheap VPSes ususally are using outdated kernels (in my case its 4.19.0) and it is not possible to use regular wireguard (as it isn't possible to load extra modules to kernel). As a workaround, it is possible to run wireguard in user namespace, i.e. [wireguard-go](https://git.zx2c4.com/wireguard-go) implementation.

There is already a [proper docker image](https://github.com/masipcat/wireguard-go-docker) which allow to run wireguard in user namespace, but it is lacking nice UI.

By combine of `wireguard-go-docker` and `wg-easy`, it is possible to create simple VPN even on cheap VPSes.

Example dockerfile:

```
FROM masipcat/wireguard-go:0.0.20220316 as wireguard-go

FROM weejewel/wg-easy:latest

# Copy wireaguard-go and wg* binaries - this will replace standard binaries, which depend on kernel module
COPY --from=wireguard-go /usr/bin/wireguard-go /usr/bin/wg* /usr/bin/
```

I tested it on 2 different VPSes and it seems to work just fine - of course performance will be worse than with native kernel support, but its fast enough saturate my 100Mb/s connection (which isn't a case for i.e. open-vpn), so it should be enough for home users which want to use VPN to their own VPS.

Is it possible, that you will add extra tag to `wg-easy` which will use wireguard-go implementation (based on example above)?

archont94:
Resolves WeeJeWel/wg-easy#327

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -62,13 +62,21 @@ module.exports = class WireGuard {
         await Util.exec('wg-quick down wg0').catch(() => { });
         await Util.exec('wg-quick up wg0').catch(err => {
           if (err && err.message && err.message.includes('Cannot find device "wg0"')) {
-            throw new Error('WireGuard exited with the error: Cannot find device "wg0"\nThis usually means that your host\'s kernel does not support WireGuard! Trying to use wireguard-go implementation.');
-            await Util.exec('ln -s /wireaguard/wireguard-go /usr/local/sbin/wireguard-go').catch(() => { });
-            await Util.exec('ln -s /wireaguard/wg /usr/local/sbin/wg').catch(() => { });
-            await Util.exec('ln -s /wireaguard/wg-quick /usr/local/sbin/wg-quick').catch(() => { });
-          }
+            debug('WireGuard exited with the error: Cannot find device "wg0"\nThis usually means that your host\'s kernel does not support WireGuard! Trying to use wireguard-go implementation.');
+            Util.exec('mkdir -p /usr/local/sbin').catch(() => { });
+	    Util.exec('ln -s /wireaguard-go/wireguard-go /usr/local/sbin/wireguard-go').catch(() => { });
+            Util.exec('ln -s /wireaguard-go/wg /usr/local/sbin/wg').catch(() => { });
+            Util.exec('ln -s /wireaguard-go/wg-quick /usr/local/sbin/wg-quick').catch(() => { });
 
-          throw err;
+	    Util.exec('wg-quick up wg0').catch(err => {
+              if (err && err.message && err.message.includes('Cannot find device "wg0"')) {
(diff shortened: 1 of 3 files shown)


src/lib/WireGuard.js
```diff
@@ -62,10 +62,19 @@ module.exports = class WireGuard {
         await Util.exec('wg-quick down wg0').catch(() => { });
         await Util.exec('wg-quick up wg0').catch(err => {
           if (err && err.message && err.message.includes('Cannot find device "wg0"')) {
-            throw new Error('WireGuard exited with the error: Cannot find device "wg0"\nThis usually means that your host\'s kernel does not support WireGuard!');
+            debug('WireGuard exited with the error: Cannot find device "wg0"\nThis usually means that your host\'s kernel does not support WireGuard! Trying to use wireguard-go implementation.');
+            Util.exec('cp -f /wireguard-go/wg /wireguard-go/wg-quick /wireguard-go/wireguard-go /usr/bin/').catch(() => { });
+            Util.exec('wg-quick up wg0').catch(err => {
+              if (err && err.message && err.message.includes('Cannot find device "wg0"')) {
+                throw new Error('WireGuard exited with the error: Cannot find device "wg0"\nThis means, that even wireguard-go does not work!\nThis usually means that you didn\'t mounted /dev/net/tun to container properly!');
```
eandersons, line 69: Grammar: `didn't mounted` > `didn't mount`.

eandersons (changes requested):

WeeJeWel:
Would be better to download & make `wireguard-go` ourselves instead of relying on someone's docker image.

archont94:
> Would be better to download & make `wireguard-go` ourselves instead of relying on someone's docker image.

External dependencies removed.

gaby:
@archont94 @WeeJeWel  Alpine already publishes an official package for `wireguard-go`. It's in the edge tag.

```
FROM alpine:edge

RUN apk update && apk add wireguard-go
```

Package link: https://pkgs.alpinelinux.org/package/edge/testing/x86_64/wireguard-go

rudasn:
> @archont94 @WeeJeWel Alpine already publishes an official package for `wireguard-go`. It's in the edge tag.
> 
> ```
> FROM alpine:edge
> 
> RUN apk update && apk add wireguard-go
> ```
> 
> Package link: https://pkgs.alpinelinux.org/package/edge/testing/x86_64/wireguard-go

I'm not familiar with alpine, how stable is the edge tag? Is there a time frame for when wireguard-go will be included in latest?

pheiduck:
A better solution is build here: https://github.com/wg-easy/wg-easy/pull/702
This can be closed.

PR #331, opened 2022-07-28
Add boringtun support

zodiacg:
* Add builtin boringtun binary from leonnicolas/boringtun:alpine.
* Activate boringtun by setting environment variable and mounting device. Description has added to the README.md

It might be nicer to build the boringtun support into a seperated tag to avoid the 3MB if not used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,8 @@ This image has builtin [boringtun](https://github.com/cloudflare/boringtun) supp
 [boringtun](https://github.com/cloudflare/boringtun) is a userspace wireguard implementation using Rust. It is similar to [wireguard-go](https://git.zx2c4.com/wireguard-go) with a smaller memory footprint.
 The TUN device `/dev/net/tun` needs to exist on the host for `boringtun` to work.
 
-To activate `boringtun` in wg-easy, you need to set the `WG_QUICK_USERSPACE_IMPLEMENTATION` environment variable and mount the `/dev/net/tun` device when running the container.
+To activate `boringtun` in wg-easy, you need to mount the `/dev/net/tun` device when running the container.
+The `wg-quick` will automatically fallback to userspace implementation if the kernel module doesn't exist.
 
 <pre>
 $ docker run -d \
@@ -93,11 +94,8 @@ $ docker run -d \
   --sysctl="net.ipv4.conf.all.src_valid_mark=1" \
   --sysctl="net.ipv4.ip_forward=1" \
   --restart unless-stopped \
-  -e WG_QUICK_USERSPACE_IMPLEMENTATION=boringtun
   weejewel/wg-easy
 </pre>
-
-Big thanks to leonnicolas for his [boringtun docker image](https://github.com/leonnicolas/boringtun).
 
 ### 3. Sponsor
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,6 @@ $ docker run -d \
   -e WG_HOST=<b>🚨YOUR_SERVER_IP</b> \
   -e PASSWORD=<b>🚨YOUR_ADMIN_PASSWORD</b> \
   -v ~/.wg-easy:/etc/wireguard \
-  -device /dev/net/tun:/dev/net/tun \
   -p 51820:51820/udp \
   -p 51821:51821/tcp \
   --cap-add=NET_ADMIN \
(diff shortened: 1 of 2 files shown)


archont94:
[This benchmark](https://github.com/cloudflare/boringtun/issues/102#issuecomment-610459038) shows impressive boringtun performance, very close to kernel wireguard. In that case, I think #327 can be closed and boringtun should be implemented instead. 

@zodiacg you can try to copy [my fallback code](https://github.com/WeeJeWel/wg-easy/pull/328/files#diff-65b7aac2a6e96abf37ec0c4c0c24a4e94ac70fc6a0702d74f45f79b900716f79R64-R76), to have automatic switch to boringtun if kernel module isn't available.

zodiacg:
> @zodiacg you can try to copy [my fallback code](https://github.com/WeeJeWel/wg-easy/pull/328/files#diff-65b7aac2a6e96abf37ec0c4c0c24a4e94ac70fc6a0702d74f45f79b900716f79R64-R76), to have automatic switch to boringtun if kernel module isn't available.

Yes. I think config-free can be done by adding those environment variables to activate boringtun after a failed initialization. I'll look into that when I got time. It has been years since last time I wrote js XD

README.md
```diff
@@ -71,6 +71,34 @@ The Web UI will now be available on `http://0.0.0.0:51821`.
...
+  -v ~/.wg-easy:/etc/wireguard \
+  -device /dev/net/tun:/dev/net/tun \
+  -p 51820:51820/udp \
+  -p 51821:51821/tcp \
+  --cap-add=NET_ADMIN \
+  --cap-add=SYS_MODULE \
+  --sysctl="net.ipv4.conf.all.src_valid_mark=1" \
+  --sysctl="net.ipv4.ip_forward=1" \
+  --restart unless-stopped \
+  -e WG_QUICK_USERSPACE_IMPLEMENTATION=boringtun
```
archont94, line 96: Can be removed, as already set in Dockerfile.

README.md
```diff
@@ -71,6 +71,34 @@ The Web UI will now be available on `http://0.0.0.0:51821`.
 
 > 💡 Your configuration files will be saved in `~/.wg-easy`
 
+#### 2.1 Using boringtun
+
+This image has builtin [boringtun](https://github.com/cloudflare/boringtun) support for machines with older kernel or limited virtual environments such as OpenVZ.
+[boringtun](https://github.com/cloudflare/boringtun) is a userspace wireguard implementation using Rust. It is similar to [wireguard-go](https://git.zx2c4.com/wireguard-go) with a smaller memory footprint.
+The TUN device `/dev/net/tun` needs to exist on the host for `boringtun` to work.
+
+To activate `boringtun` in wg-easy, you need to set the `WG_QUICK_USERSPACE_IMPLEMENTATION` environment variable and mount the `/dev/net/tun` device when running the container.
```
archont94, line 80: Can be removed, as already set in Dockerfile.

zodiacg:
It's probably ok now. I just noticed wg-quick will only use userspace implementations (wireguard-go or boringtun) if the kernel module doesn't exist. So it's ok to just set the env variables and leave the web part as-is.

I've done some quick tests. On a dedicated server (with wireguard kmod available), there is no boringtun process. While on an OpenVZ7 VPS, the boringtun appears and almost everything seems fine.

However I did notice some trouble regarding config files. I forgot to take a screenshot or copy the logs. But I have to wipe the config files everytime I test a newer image I built, or there will be errors to prevent it from working.

ethack:
> But I have to wipe the config files everytime I test a newer image I built, or there will be errors to prevent it from working.

I ran into this too, but the problem was present when adding a second peer. It would also happen if trying to start wg-easy with an existing config. E.g. restart the docker container.

Here was my error:
```
  Server Listening on http://0.0.0.0:51821 +0ms
  WireGuard Loading configuration... +0ms
  WireGuard Configuration loaded. +1ms
  WireGuard Config saving... +1ms
  WireGuard Config saved. +1ms
$ wg-quick down wg0
$ wg-quick up wg0
  WireGuard Config syncing... +155ms
$ wg syncconf wg0 <(wg-quick strip wg0)
Error: Command failed: wg syncconf wg0 <(wg-quick strip wg0)
Unable to modify interface: Protocol error

    at ChildProcess.exithandler (child_process.js:383:12)
    at ChildProcess.emit (events.js:400:28)
    at maybeClose (internal/child_process.js:1058:16)
    at Socket.<anonymous> (internal/child_process.js:443:11)
    at Socket.emit (events.js:400:28)
    at Pipe.<anonymous> (net.js:686:12) {
  killed: false,
  code: 1,
  signal: null,
  cmd: 'wg syncconf wg0 <(wg-quick strip wg0)'
}
```

I `exec`ed into the container and determined this was due to boringtun hanging. The boringtun process would still be running but, `wg show` would also hang. Starting a new `boringtun wg0` process allowed it to work again.

I'm pretty sure https://github.com/cloudflare/boringtun/issues/40 is the culprit here, which is triggered by wg-easy's use of `wg syncconf`. I changed [this line](https://github.com/WeeJeWel/wg-easy/blob/736c62b4369dc7a6502139a3cbd3a177d3f61727/src/lib/WireGuard.js#L129) to use `wg setconf` instead and my error went away.

I'm assuming there's a downside of using `setconf` vs `syncconf`. My guess is that established client connections will get reset, but I haven't tested that. This could still be a valid workaround as long as clients are infrequently changed.

ruimarinho:
@zodiacg I see your version is using `wg syncconf` - have you not ran into the issues described by @ethack?

zodiacg:
@ruimarinho Yes I have. I think that's exactly the reason behind the problem I described [above](https://github.com/WeeJeWel/wg-easy/pull/331#issuecomment-1198380277). I made another [branch](https://github.com/zodiacg/wg-easy/tree/synconf) to use [ernestask's modification](https://github.com/ernestask/boringtun/tree/wip/ernestask/modify-peers), which allows `wg syncconf`.

TermeHansen:
This seems very nice, any reason why this is not merged in? 
@zodiacg should I just build a docker image myself following your [branch](https://github.com/zodiacg/wg-easy/tree/synconf) ?

zodiacg:
> This seems very nice, any reason why this is not merged in? @zodiacg should I just build a docker image myself following your [branch](https://github.com/zodiacg/wg-easy/tree/synconf) ?

Yes, I think you can give it a try. Though personally I haven't worked on this for a while since I no longer use any OpenVZ based VPS.

TermeHansen:
thks @zodiacg , it is running now on my OpenVZ instance, only thing I needed to change was to bump rust. I just bumped to 1.63 and it compiled again:

https://github.com/zodiacg/wg-easy/blob/75bf8d1c812c0bd00debaca0af40ffa8708f1f03/Dockerfile#L13

to 

`FROM rust:1.63 AS boringtun-builder
`

PR #332, opened 2022-07-29
Document option to change the default WebUI port

danifr:
The option is there but not documented https://github.com/WeeJeWel/wg-easy/blob/master/src/config.js#L6


WeeJeWel:
You must change the docker mapping.

PR #341, opened 2022-08-03
Import json configurations with no PreShared-Key

t-oster:
Hi. When migrating old setups, one may want to keep existing clients, so you don't have to roll out all new configs.

In our case our clients do not yet have a PSK, but if we remove them from the config, wg-easy won't accept it and die.

This patch just changes the config-generation so that missing PSKs will just be ignored. The UI and the generation of new users is not touched, so this only affects configs, where you manually remove the PSK from the wg0.json.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -110,8 +110,8 @@ PostDown = ${WG_POST_DOWN}
 # Client: ${client.name} (${clientId})
 [Peer]
 PublicKey = ${client.publicKey}
-PresharedKey = ${client.preSharedKey}
-AllowedIPs = ${client.address}/32`;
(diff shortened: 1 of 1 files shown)


qy2009:
I would love this option also, my old client doesn't support pre-shared key :(

segalws:
Me need too

dvision1979:
Any updates on this one. I am also interested in this feature. It is a real-world scenario, when migrating from other setups.
Thanks!

t-oster:
@dvision1979 could you tell me if there is anything wrong with my request or just no interest in the feature?

pheiduck:
@t-oster please rebase.

t-oster:
rebased and linted.

PR #20, opened 2021-05-30
add license and fix iptables issues

ragnarok22:

PR #342, opened 2022-08-04
Truly configure wireguard port from WG_PORT envvar

dangra:


eandersons:
There have already been similar pull requests: #286, #285, #257, #87, #20 and partly #81, and they all have been rejected because:

> `WG_PORT` is only used for the clients information. Docker maps this port.

dangra:
Ah, got your point. It doesn't work for me because I am running it in fly.io but can make it work with my fork.
Would you accept a PR to add WG_INTERNAL_PORT instead?

hacker-hackman:
@dangra could you please help me? :-)
I just can't understand how to bind to `fly-global-services` #1782

PR #394, opened 2022-10-10
Allow running this software as unprivilidged software

BrainStone:
This allows running the software as an unpriviledged user.  
All that user needs to be allowed to run these commands without password:
- `wg`
- `wg-quick`

This can be achieved by putting this in the sudoers config:

```sudo
<user>       ALL=(root)      NOPASSWD: /usr/bin/wg , /usr/bin/wg-quick
```

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -2,6 +2,8 @@
 
 const { release } = require('./package.json');
 
+module.exports.USE_SUDO = process.env.USE_SUDO || (process.getuid() !== 0);
+module.exports.SUDO_STRING = module.exports.USE_SUDO ? 'sudo ' : '';
 module.exports.RELEASE = release;
 module.exports.PORT = process.env.PORT || 51821;
 module.exports.PASSWORD = process.env.PASSWORD;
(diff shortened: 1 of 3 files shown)


rudasn:
Maybe this could be a simpler change, and more likely to be accepted, if instead you allowed the `wg` and `wg-quick` commands to be provided via environment variables instead. For example, in docker compose pass the environment variables `WG_CMD=sudo wg` and `WG_QUICK_CMD=sudo wg-quick`.

Nice work, I'd also like to see this get merged.

BrainStone:
I don't think it makes sense to abstract it to that level. At least with the intentions of my PR.

WeeJeWel:
Why is this better? It runs inside a Docker container.

BrainStone:
> Why is this better? It runs inside a Docker container.

What if it doesn't? Just because a docker container exists, doesn't mean it shouldn't be able to be ran directly.

Additionally it's an unnecessary security risk to run containers with a root user inside. Always go with least priviledges.

PR #395, opened 2022-10-11
Allow configuring of bind/listen IP

BrainStone:
Always a good idea to be able to prevent a server from binding 0.0.0.0


rahilarious:
Seems like I duplicated efforts in https://github.com/wg-easy/wg-easy/pull/566

PR #313, opened 2022-07-09
add allowed IPs to client creation

dcalvo:
This PR adds the functionality described in #243. I needed this for my own use and it was easy enough to share. I added an additional text box to the Create Client window, the value of which is forwarded to the `CreateClient` route. 

I tweaked template generation to make this work, and also slightly changed the next IP calculation to make it more robust in certain situations (i.e. detecting `10.8.0.5/32, 10.8.0.6/32, 10.8.0.7/32` as three occupied IPs).

Vue is not my strong suit so please feel free to redesign as you wish.

PR #400, opened 2022-10-17
Add Allowed IPs on Gateway Peers

wingsman2:
This PR is a modification of #313 to expand upon adding allowed IPs. I put another variable in wg0.json instead of using the address variable. I also created a second edit field next to the edit address section on the clients to allow you to update or modify allowed IPs.

I am not super amazing at programing so let me know if you see any mistakes.


megoRU:
Request: 
```JSON
{
"name": "test", 
"allowedGWIPs": "10.8.1.29"
}
```
```JSON
{
  "error": "Maximum number of clients reached.",
  "stack": "Error: Maximum number of clients reached.\n at WireGuard.createClient (/app/lib/WireGuard.js:254:13)\n at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
}
```

`10.8.0.254`
if you take all 255 addresses

wingsman2:
> Request:
> 
> ```json
> {
> "name": "test", 
> "allowedGWIPs": "10.8.1.29"
> }
> ```
> 
> ```json
> {
>   "error": "Maximum number of clients reached.",
>   "stack": "Error: Maximum number of clients reached.\n at WireGuard.createClient (/app/lib/WireGuard.js:254:13)\n at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
> }
> ```
> 
> `10.8.0.254` if you take all 255 addresses

In the allowed IP section did you put in full CIDR? I don't have any error detection because this was a simple add, but you would need to put in the Allowes IPs as 10.8.0.29/24 or what ever subnet type you want

megoRU:
> 

I also tried `10.8.1.29/32`, but for some reason it added 2 IP addresses at once (`10.8.0.29`).  and in the end, an error was caused that is not available for creation due to the fact that the limit is exceeded (255)

megoRU:
And I thought your code fixes something else. Sorry)

jlcs-es:
Hi, this featue would be very handy for my usecase: I want to access a client's subnet.
For this, I need the server's wg config for the peer to look someting like:

```
[Peer]
PublicKey = mypub
PresharedKey = mykey
AllowedIPs = 10.8.0.5/32, 192.168.0.0/24
```


What is the state of this PR?
As I understand, there are edge cases with the gui which are not handled.

**Would it be acceptable to only include the changes to `src/lib/WireGuard.js` related to the config JSON file and the `wg0.conf` generation?** This way, "advanced" users could make the change to the JSON file and have it applied in wireguard, while the really basic users would limit themselves to the GUI and not break anything.

Note: also maybe instead of calling the json attribute `allowedGWIPs`, we should call it `extraAllowedIPGWs`, to signify that the `client.address` + /32 is always included.

wingsman2:
> Hi, this featue would be very handy for my usecase: I want to access a client's subnet. For this, I need the server's wg config for the peer to look someting like:
> 
> ```
> [Peer]
> PublicKey = mypub
> PresharedKey = mykey
> AllowedIPs = 10.8.0.5/32, 192.168.0.0/24
> ```
> 
> What is the state of this PR? As I understand, there are edge cases with the gui which are not handled.
> 
> **Would it be acceptable to only include the changes to `src/lib/WireGuard.js` related to the config JSON file and the `wg0.conf` generation?** This way, "advanced" users could make the change to the JSON file and have it applied in wireguard, while the really basic users would limit themselves to the GUI and not break anything.
> 
> Note: also maybe instead of calling the json attribute `allowedGWIPs`, we should call it `extraAllowedIPGWs`, to signify that the `client.address` + /32 is always included.

If you only updated the WireGuard.js You wouldn't be able to add to the "AllowedIPs" section because it requires either the fill in section on the main page or the text box when you create a new client.

image
image
image

jlcs-es:
> If you only updated the WireGuard.js You wouldn't be able to add to the "AllowedIPs" section because it requires either the fill in section on the main page or the text box when you create a new client.

Not necessarily. You could do:
Create de Peer -> Modify the `wg0.json` file -> Reload wg-easy.

PR #407, opened 2022-10-27
Add PASSWORD_FILE environmental variable support

### LINKED ISSUE #401: [feature request] PASSWORD_FILE instead of PASSWORD
**Describe the feature**
Add the possibility to load the `PASSWORD` from a mounted `docker` or `podman` secret.

Currently the only way to set the `PASSWORD` is to define it as a environmental variable. This is not considered secure, secrets should (and by default often are) defined as files mounted in the running container.

`wg-easy` should still allow for the `PASSWORD` environmental variable to be defined, but should also have the possibility to define a `PASSWORD_FILE` environmental variable, which would hold the path to the mounted file containing the secret. If both are defined it should either error out or prefer the `PASSWORD_FILE` as the more secure of the two.

czocher:
When a PASSWORD_FILE environmental variable is provided read the PASSWORD from it instead of using the PASSWORD environmental variable directly.

Closes #401.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ These options can be configured by setting environment variables using `-e KEY="
 | Env | Default | Example | Description |
 | - | - | - | - |
 | `PASSWORD` | - | `foobar123` | When set, requires a password when logging in to the Web UI. |
+| `PASSWORD_FILE` | - | `/run/secrets/password` | Path to a mounted password file. When set, contents of the file will override value from `PASSWORD`. |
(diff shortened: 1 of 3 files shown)


jmigual:
Why is this still not merged? Not using a PASSWORD_FILE is what's preventing me from using this.

czocher:
@jmigual this project is rather dead. I used a forked version myself with this PR included. Feel free to do the same. I really think a hard fork or rewrite would be beneficial though (I'm strongly considering it ATM and found some friends who are willing to help me).

jmigual:
What kind of joke is this?

PR #413, opened 2022-11-02
Added simple sortation feature

eftas-gka:
Client list is now sorted by client name by default. I also added a control to change sortation back to address.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /config
 /wg0.conf
 /wg0.json
+/.idea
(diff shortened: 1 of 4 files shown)


rudasn:
Nice! Could you add a screenshot of how it looks?

Have you thought about adding more sorting options, like current status (eg. online, offline, disabled).

WeeJeWel:
Can you share a screenshot?

peterlewis:
Echoing @WeeJeWel's comment that a screenshot of this implemented would be greatly appreciated. Thanks!

pheiduck:
Well we got no reply from author 🤷‍♂️

PR #447, opened 2023-01-03
clients traffic statistic (downloaded\uploaded)

FlestRus:
it would be nice to have statistics)


hadi-dindar:
how we can add this  statistics  to pannel ?

oplexz:
You can already see the Download/Upload statistics by hovering over client's current download/upload speeds.

PR #502, opened 2023-02-25
User-defined ListenPort of the WireGuard interface

qMalte:
The port set by the environment variables is not set correctly as ListenPort of the WireGuard interface. So far, this is only used for the client configurations.

In the file: /src/lib/WireGuard.js the port is hardcoded in line 98. However, the variable WG_PORT should also be used here.

This PR fixes the error mentioned above, so that the configuration of a user-defined port is now possible.


n-hass:
Merging this makes sense. I see the value in hard-coding it, but it becomes less transparent and maintainable for users as this isn't quite the norm and isn't the first logical conclusion you'd make when reading the environment var "WG_PORT".

It also means the container cant be run in host network mode if there are conflicting port assignments on the default 51820.

I can't see a configuration in which making this change becomes a serious complication or regression, but the use case above ^ is not that uncommon, yet its currently impossible with the hardcoded port.

PR #522, opened 2023-04-01
Patch to make custom DNS work via WG_DEFAULT_DNS

m31u:
Currently `WG_DEFAULT_DNS` is only added to the peer config, but is required in the main config for DNS. Without this change defining the `WG_DEFAULT_DNS` environment variable will cause the peer's DNS lookups to fail to resolve.

### DIFF
--- a/src/lib/WireGuard.js
+++ b/src/lib/WireGuard.js
@@ -100,6 +100,7 @@ PreUp = ${WG_PRE_UP}
 PostUp = ${WG_POST_UP}
 PreDown = ${WG_PRE_DOWN}
 PostDown = ${WG_POST_DOWN}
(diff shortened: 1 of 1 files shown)


rahilarious:
> Without this change defining the `WG_DEFAULT_DNS` environment variable will cause the peer's DNS lookups to fail to resolve.
It DOES work without this patch. 

> Currently WG_DEFAULT_DNS is only added to the peer config, but is required in the main config for DNS
Setting DNS in [Interface] of server config is redundant. I can't think of any scenario where it is needed but yeah it shouldn't harm to set it.

WeeJeWel:
It’s not specified in the wg manual so this is not going to do anything.

PR #536, opened 2023-04-21
Allow use WG_HOST=auto to resolve host programatically

barsikus007:
This PR allows to use auto value of WG_HOST like in other docker wireguard implementation ([linuxserver/docker-wireguard](https://github.com/linuxserver/docker-wireguard/blob/ac5f805be80e558ccec1492e90295d1a179739b4/root/etc/s6-overlay/s6-rc.d/init-wireguard-confs/run#L145)) to resolve IP of server for configs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ RUN npm i -g nodemon
 
 # Install Linux packages
 RUN apk add -U --no-cache \
+  curl \
   wireguard-tools \
   dumb-init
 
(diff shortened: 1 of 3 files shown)


rahilarious:
nice Quality of Life PR. Should be merged.

WeeJeWel:
Thanks, but I don’t like the dependency on someone else’s server. I think it’s also better to specify the hostname so the user is aware it’s static enough and ports are forwarded.

PR #548, opened 2023-05-09
Add PORT environment variable in README.md

rahilarious:


WeeJeWel:
It’s just more confusing because Docker maps the port already.

rahilarious:
What if user is using `--network=host` and want webUI on other port ?

WeeJeWel:
Then they set the PORT environment variable :)

rahilarious:
and how would they know that env is configurable when its absent in docs?

WeeJeWel:
﻿The docs are for the Docker usecase only. If they want to be adventurous, they can dive into the source files.Sent from my iPhoneOn 17 May 2023, at 10:51, Rahil Bhimjiani ***@***.***> wrote:﻿
and how would they know that env is configurable when its absent in docs?

rahilarious:
It is an official docker-cli command switch but Your project your rules. Respecting that. 

Thanks for minimal & pretty project.

WeeJeWel:
You can always write a wiki article for running Docker in host mode :)

rahilarious:
That would be the shortest article EVER ;)

PR #558, opened 2023-05-24
Make PresharedKey optional

IJMacD:
(if editing config manually)

When importing some clients I needed to add clients which don't have a Pre-Shared Key set.

This change generates correct wg conf if no pre-shared key is set.


IJMacD:
Duplicate of #341 

Sorry this is essentially a duplicate. I didn't find the other PR before submitting mine.

PR #590, opened 2023-07-01
Changed changelog.json URL

rainloreley:
It seems like this repository was transferred from github/weejewel to github/wg-easy before, but the URL of the changelog file hasn't been changed, so the wg-easy website would fail to fetch and display the newest version and throw an 404 error in the background.

grafik

https://github.com/wg-easy/wg-easy/blob/642d6d62edb95a96e57b704966b65135f4f3ea94/src/www/js/app.js#L273

This PR changes the changelog URL to `https://wg-easy.github.io/wg-easy/changelog.json`

### DIFF
--- a/src/www/js/app.js
+++ b/src/www/js/app.js
@@ -270,7 +270,7 @@ new Vue({
 
     Promise.resolve().then(async () => {
       const currentRelease = await this.api.getRelease();
-      const latestRelease = await fetch('https://weejewel.github.io/wg-easy/changelog.json')
+      const latestRelease = await fetch('https://wg-easy.github.io/wg-easy/changelog.json')
         .then(res => res.json())
         .then(releases => {
           const releasesArray = Object.entries(releases).map(([version, changelog]) => ({


amirdana:
So basically all forks are now out of date and problematic, nice.

rainloreley:
> So basically all forks are now out of date and problematic, nice.

@amirdana I don't quite get what you're trying to say. Fork maintainers can just sync their repository once this PR is merged. Apart from that, it's not a good idea anyways to pull the changelog from the upstream repo if you're running a forked version. Having an update available from this repo doesn't mean that there's one available immediately for the forked repo if they stay on one version or if they're adding additional features. It's better to check for updates via Docker, services like Portainer can show you available updates via their dashboard.

It's just a simple bug which can be synced to existing forks if their maintainers wish to do so

peterlewis:
Thank you. This has been fixed in https://github.com/wg-easy/wg-easy/commit/a89ff96d37706c5e8c38a4b1297d3c58b286b56b.

PR #597, opened 2023-07-12
sessionless auth

EugeneDoingThings:


peterlewis:
Thank you so much for your contribution @EugeneDoingThings. This feature is to be added in version 9 (which is currently in `nightly`).

PR #623, opened 2023-09-13
Avoid hardcoded wg port value in iptables rule

pitabwire:
Avoids hard coded wireguard port when setting the ip tables port value but instead utilizes the configuration value.


eandersons:
Did you take a look at pull requests before submitting your own? There already are [pull requests (both opened and closed) like yours](/wg-easy/wg-easy/pulls?q=is%3Apr+wg_port).

PR #638, opened 2023-10-07
Create wg-easy-npm.yml

comfuzio:
I am running portainer already with NPM as reverse proxy for websites that I am hosting. So I have modded this yml with the help of ChMaikos. You will need to put the "wg-easy" in the IP section of your NPM and use port 51821


WeeJeWel:
Good for you, but not useful for others.

PR #639, opened 2023-10-09
Determine interface of default network rather than assume eth0

cjd:
This small change defines WG_INTERFACE based on default route interface rather than assuming eth0 (handy when interface may change)


pheiduck:
Some people use this: https://wiki.archlinux.org/title/Network_configuration#Change_interface_name to change there interface name.

Issue #595, opened 2023-07-10
Readme/Screenshot should clarify that keys are for devices (1 on 1) and not for users/humans


thomasmerz:
Due to the fact that wireguard uses keys to **authenticate a device** and not a user this should be clearly stated and shown on Readme by a screenshot with "tom's computer", "tom's laptop", "tom's smartphone-private", "tom's smartphone-work", …
See also https://serverfault.com/questions/948816/wireguard-user-authentication for details.

thomasmerz:
@WeeJeWel , could you please update the screenshot in the Readme so that it's clear that WireGuard authenticates per device and not per user/human? Thanks a lot! 😗

WeeJeWel:
You can make a PR for a better screenshot :-)

thomasmerz:
I wish I could… but unfortunately I'm using plain WireGuard without your great tool anymore… 🤷🏼‍♂️ But I can and will add some nice text to the screenshot that clarifies how WG authenticates via a PR.

thomasmerz:
Won't be fixed as nobody seems to be able to provide a screenshot that clarifies that WireGuard keys are for Emile's, Jason's, Linus' and Adam's **device** and not for the persons themself. 😞

PR #680, opened 2023-12-18
Issue 595 📝 fix Readme regarding WG uses device-auth and not user-auth

thomasmerz:
This PR fixes issue #595.


WeeJeWel:
Sorry, I don't like extra text, especially not disclaimers. A new screenshot would remove the need for this warning.

PR #687, opened 2023-12-21
Dockerfile: use  iptables-legacy and symlink to iptables

pheiduck:


peterlewis:
Would something along the lines of `update-alternatives --set iptables /usr/sbin/iptables-legacy` be more appropriate?

PR #697, opened 2023-12-22
👷 use latest major version for actions/checkout

thomasmerz:
Use latest major version of [actions/checkout](https://github.com/actions/checkout) for GitHub Actions.


pheiduck:
@thomasmerz is your fork in sync with the latest master from here?
Seems not because we already at latest version 4 of actions/checkout

thomasmerz:
You @pheiduck are totally right - I totally forgot to sync my fork. Sorry for the inconvenience… and thanks for your hint to update my forks… 😉

PR #698, opened 2023-12-22
👷 use latest major version for actions/setup-node

thomasmerz:
Use latest major version of [actions/setup-node](https://github.com/actions/setup-node) for GitHub Actions


pheiduck:
@thomasmerz is your fork in sync with the latest master from here?
Seems not because we already at latest version 4 of actions/checkout

Issue #735, opened 2024-01-06
Port Change Bug


sysvar:
**Describe the bug**
If you change the default port 51820 to any other udp port for both environment variable and exposed ports it fails to allow connections. tcpdump showed it did receive packets. GUI shows no connection stats.

**To Reproduce**
1. Change the **WG_PORT** environment variable from **51820** to **63000**
2. Change the exposed **PORTS** from **-p 51820:51820/udp** to this **-p 63000:63000/udp**

**Not Working Config**
```
version: "3.8"
volumes:
  etc_wireguard:

services:
  wg-easy:
    environment:
      # ⚠️ Required:
      # Change this to your host's public address
      - WG_HOST=<<<OMIT>>>

      # Optional:
      - WG_DEVICE=eth0
      - PASSWORD=putsomethinghere
      - WG_PORT=63000
      # - WG_DEFAULT_ADDRESS=10.8.0.x
      # - WG_DEFAULT_DNS=1.1.1.1
      # - WG_MTU=1420
      # - WG_ALLOWED_IPS=192.168.15.0/24, 10.0.1.0/24
      # - WG_PERSISTENT_KEEPALIVE=25
      #- WG_PRE_UP=echo "Pre Up" > /etc/wireguard/pre-up.txt
      #- WG_POST_UP=echo "Post Up" > /etc/wireguard/post-up.txt
      #- WG_PRE_DOWN=echo "Pre Down" > /etc/wireguard/pre-down.txt
      #- WG_POST_DOWN=echo "Post Down" > /etc/wireguard/post-down.txt

    image: ghcr.io/wg-easy/wg-easy
    container_name: wg-easy
    volumes:
      - etc_wireguard:/etc/wireguard
    ports:
      - "63000:63000/udp"
      - "51821:51821/tcp"
    restart: unless-stopped
    cap_add:
      - NET_ADMIN
      - SYS_MODULE
    sysctls:
      - net.ipv4.ip_forward=1
      - net.ipv4.conf.all.src_valid_mark=1
```

**Working Config**
```
version: "3.8"
volumes:
  etc_wireguard:

services:
  wg-easy:
    environment:
      # ⚠️ Required:
      # Change this to your host's public address
      - WG_HOST=<<<OMIT>>>

      # Optional:
      - WG_DEVICE=eth0
      - PASSWORD=putsomethinghere
      - WG_PORT=51820
      # - WG_DEFAULT_ADDRESS=10.8.0.x
[...]

PR #736, opened 2024-01-06
Variable-ise exposed port with fallback

peterlewis:


peterlewis:
This is just a proposal in response to https://github.com/wg-easy/wg-easy/issues/735. Not sure whether variable fallback will work with `docker run`. Will need to give it a whirl...

peterlewis:
Oops! Renaming branch closed this. Fixing that!

pheiduck:
I think we got a proper fix for that here: #973

PR #741, opened 2024-01-07
Add Contributing Guidelines

pheiduck:
The `contributing.md` is autogenerated 
Needs Review from @WeeJeWel and @peterlewis


peterlewis:
I like it, but it feels a little bit long! 😀

pheiduck:
> I like it, but it feels a little bit long! 😀

You are welcome to suggest a shorter form.
I would also be in favor of it being a little shorter.

PR #745, opened 2024-01-10
IPv6 support

pheiduck:


pheiduck:
needs reopen due rebase broke, redo with patch

PR #750, opened 2024-01-11
Replace Express with Koa

oplexz:
Kinda WIP since I haven't tested this thoroughly. Profile creation, enabling and editing seems to work, but I've experienced some errors while logging in.

Needs testing & code review :)

@pheiduck


pheiduck:
@oplexz Please fix lint errors by using:
```
cd src
npm ci
npm run lint -- --fix
```

oplexz:
I've done my testing, everything appears to be working fine with that last commit, which fixed the login flow.

pheiduck:
@oplexz Finally, can you describe what exactly is the background to replacing Express with Koa?

oplexz:
> @oplexz Finally, can you describe what exactly is the background to replacing `Express` with `Koa`? Only I can see: `Koa` is more up to date then `Express`.

The reasons for this PR:

- Koa is actively maintained, as compared to Express which only appears to be receiving critical updates;
- Koa is more lightweight and modular than Express;
- Koa supports `async`/`await` out of the box, which leads to cleaner, more readable code; Express 5 also supports `async`/`await`, but version 5 is still in beta, and hasn't really been updated for the past two years;
- Better support for error handling (Express is more error-prone and needs more manual error handling).

Regarding Express being error-prone -- personally, I haven't encountered that many errors while using wg-easy, so it's more about being future-proof, especially since there are new maintainers :wink:

pheiduck:
@peterlewis Can you please take a look here?

peterlewis:
Looks good on the surface and admittedly a bit nicer to look at! Let's get this tested in `development` build...

peterlewis (commented):
Looks good on the surface and admittedly a bit nicer to look at! Let's get this tested in development build...

oplexz:
> The Session is undefined
@pheiduck Oops, I might have missed that. I couldn't find anything that identifies the session in `ctx`, so for "New session" log, I just removed `ctx.session.id`.

pheiduck:
> > The Session is undefined
> 
> @pheiduck Oops, I might have missed that. I couldn't find anything that identifies the session in `ctx`, so for "New session" log, I just removed `ctx.session.id`.

Well, good that we merged it not into `master` so it needs more development. I don’t like the behaviour that the session says undefined.

PR #754, opened 2024-01-11
Upgrade to Vue3

oplexz:
My attempt to upgrade the Web UI to use Vue 3.

I haven't really had all that much experience with Vue, so I'd be very thankful if anyone can give me advice on codestyle, best practices and other Vue-specific things I might not know.

### Roadmap:

- Refactoring;
- Breaking down App.vue into components;
- More...

### Needs fixing/discussion:

- Charts: [apexcharts is way too heavy](https://github.com/apexcharts/apexcharts.js/issues/567) (the total .js size is currently ~600 kB) -- possibly use another library for charts?;
- Charts: Individual bars are way too wide;
- UI/UX: When a client is online, pulse background is white for some reason... not yet sure if it's something I've broken, or if that's a bug in master too.

### Running for development:

```bash
cd webui
npm ci
npm run build:watch
```

Then, in second shell:

```bash
npm run build
npm run serve
```

**Important:** if you're planning to run this for development, build static files firsthand (the reason is explained below)

### Known issues:

- If you try to `npm run serve` before `webui/dist` is created, Docker will create a dist folder, which *might* have `root` as it's owner, making you unable to build static files locally. Down the compose stack, delete the `dist` folder owned by root, build and up compose stack again.

Suggestions are very welcome.


oplexz:
Things that need fixing:
- Rx/Tx charts are missing
- Last activity time is displayed raw
- Gravatars are not working

radusuciu:
Here's are a few drive-by comments:
- I would look into using `<script setup>`
- `App.vue` is a bit too large IMO - you could factor our some components/composables probably
- you have a few functions in the composables directory. To me these seem more like utility functions that true composables since they don't deal with state
- Vue 3 is often paired with TypeScript - IMO this is great productivity boost. For a project this size you can get away without it though and I can understand why you might hesitate to add another dependency.

webui/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
```
cany748, line 2: I think we shouldn't specify the lang attribute because the application is multilingual

webui/vite.config.js
```diff
@@ -0,0 +1,26 @@
...
+  plugins: [
+    vue(),
+  ],
+  resolve: {
+    alias: {
+      '@': fileURLToPath(new URL('./src', import.meta.url))
+    }
+  },
+  build: {
+    outDir: 'dist',
```
cany748, line 17: this is the default value in vite, this line can be removed

webui/vite.config.js
cany748, line 1: You don't have to use TypeScript to build an application. However, it is strongly recommended to use the .ts extension for the .config file.

webui/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "apexcharts": "^3.45.1",
+    "crypto-js": "^4.2.0",
+    "vue": "^3.3.11",
+    "vue3-apexcharts": "^1.4.4"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-vue": "^4.5.2",
+    "autoprefixer": "^10.4.16",
+    "postcss": "^8.4.33",
+    "tailwindcss": "^3.4.1",
```
cany748, line 22: We can switch from tailwind to unocss. It works great with Vue and Vite, has more functionality and higher performance, it is also easy to configure and is fully compatible with tailwind, so you don’t have to rewrite anything.

We can replace Autoprefixer and postcss with lightningcss.

cany748:
I can help you migrate everything to <script setup> and Typescript.

PR #763, opened 2024-01-13
Import json configurations with no PreShared-Key

pheiduck:
Import json configurations with no PreShared-Key (#341) from @t-oster
new Branch for testing and merge into `master`


pheiduck:
@t-oster please test: `ghcr.io/wg-easy/wg-easy:development`

pheiduck:
hold back for version 12

t-oster:
Works for me. But I had to 'npm install bcrypt' before, but this is also true for master and may be because I am not using the dockerized version.

willtho89:
Can also confirm this working. New creation, and export of clients is working as expected. 
Importing existing configs also works and I'm able to connect without the psk.

PR #770, opened 2024-01-18
tailwind recommends installing tailwind css as PostCSS plugin

pheiduck:
see: https://tailwindcss.com/docs/installation/using-postcss
`cdn.tailwindcss.com should not be used in production. To use Tailwind CSS in production, install it as a PostCSS plugin or use the Tailwind CLI: https://tailwindcss.com/docs/installation`


pheiduck:
Until now it's just a warning in the debug konsole (Browser)

PR #774, opened 2024-01-19
update: app.css to match latest tailwindcss version

pheiduck:
in the future npm-update-bot will handle it for us.


pheiduck:
tested working as expected.

PR #775, opened 2024-01-19
index.html: use app.css

pheiduck:
tested it is now build local and not from the cdn playground


peterlewis:
Thank you, this is great.

PR #776, opened 2024-01-19
Support for semantic versioning

pheiduck:
https://semver.org/


pheiduck:
hm, when the build is published, the ".1" is cut off :/

pheiduck:
@peterlewis I think we could change that a little in a different way:
Release: "10"
version: "10.1"

so that relase always uses the major version and the version is changed when patches are applied?
What do you think?

pheiduck:
And doring this playaround tag 10 got updated what should not be happen...
I rebuild latest on production again

peterlewis:
> so that relase always uses the major version and the version is changed when patches are applied? What do you think?

Ideally we want to be tagging both, so the latest (next) release would have `latest`, `10` and `10.1`. However, it`s worth noting that the current release will not have the necessary code to pass the minor version. It`s therefore my understanding that in order to achieve this we would need to go to v11 and then be able to release minor version releases from there-on-in.

pheiduck:
Okay, then we should go to v11 with the upcoming pr #782

kaaax0815:
@pheiduck please consider implementing this. to current version scheme doesnt transfer any information

PR #793, opened 2024-01-27
allow clients with no privateKey

willtho89:
see https://github.com/wg-easy/wg-easy/discussions/791

I disabled the download from the frontend.
If someone tries to manually call the API, the privateKey block is set to "REPLACE_ME" to indicate that this value must be set manually.

As there was no class for disabled Buttons yet, i just set the opacity to 0.25:
image
image


pheiduck:
@willtho89 please fix lint errors

willtho89:
@pheiduck done

peterlewis:
Thank you! As you mentioned, there currently does not exist a class for disabled buttons. Are you happy to please add one? There's a [StackOverflow](https://stackoverflow.com/questions/64962149/tailwindcss-disabled-variant-not-working/64962285#64962285) that covers this, if it is of any help.

Thanks!

willtho89:
@peterlewis also done. Never worked with tailwind before, so hopefully this approach is okay. The biggest hurdle is that <a> tags don't have a disabled state, so we have to go this way

src/www/css/app.css
```diff
@@ -1350,6 +1350,11 @@ video {
   transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
 }
 
+.is-disabled {
```
peterlewis, line 1353: I'm guessing you have named this class `is-disabled` to differenciate from the `disabled` state? I like your thinking and think we can build on this. Would you be happy to rename to `inactive`? Allows, then, for potential wider uses! :)

src/www/index.html
```diff
@@ -225,9 +230,16 @@ <h2 class="text-sm text-gray-400 dark:text-neutral-400 mb-10"></h2>
...
-                    <a :href="'./api/wireguard/client/' + client.id + '/configuration'" download
-                      class="align-middle inline-block bg-gray-100 dark:bg-neutral-600 dark:text-neutral-300 hover:bg-red-800 dark:hover:bg-red-800 hover:text-white dark:hover:text-white p-2 rounded transition"
-                      :title="$t('downloadConfig')">
+                    <a :disabled="!client.downloadableConfig"
+                      :href="'./api/wireguard/client/' + client.id + '/configuration'"
+                      :download="client.downloadableConfig ? 'configuration' : null"
+                      class="align-middle inline-block bg-gray-100 dark:bg-neutral-600 dark:text-neutral-300 p-2 rounded transition"
+                      :class="{
+                        'hover:bg-red-800 dark:hover:bg-red-800 hover:text-white dark:hover:text-white': client.downloadableConfig,
+                        'is-disabled': !client.downloadableConfig
```
peterlewis, line 239: https://github.com/wg-easy/wg-easy/pull/793/files#r1471851844

pheiduck:
@willtho89 please fix lint errors.

willtho89:
@pheiduck done

PR #814, opened 2024-02-04
prereqs installation

pheiduck:
- CentOS 7/8/9 ✅
- Arch Linux and based on Distros ✅


pheiduck:
Hm, I might want to have this on the Wiki page, due the README will get to long

PR #831, opened 2024-02-12
Add Spanish language

amarqz:
Added Spanish localisation.


pheiduck (approved):
As we introduce multilanguage support in the Pending release, this will be merged.

PR #843, opened 2024-02-17
Fix detailed stats

suxscribe:
- Changed layout so controls are wrapped first. Info block and RX/TX stay in line until 450px width
- Fix last seen showing `54 years` on new clients


pheiduck:
@suxscribe some issues:
```
SyntaxError: JSON.parse: expected ',' or '}' after property value in object at line 12 column 3 of the JSON data
```
```
ReferenceError: uiDetailedStats is not defined
```

suxscribe:
@pheiduck Got it!

PR #844, opened 2024-02-17
Fix detailed stats 2

suxscribe:
- Fixed error in changelog.json
- Updated css
- Some refactoring on fetching uiDetailedStats

Hope everything's ok this time :)

### DIFF
--- a/docs/changelog.json
+++ b/docs/changelog.json
@@ -8,6 +8,6 @@
   "7": "Improved the look & performance of the upload/download chart.",
   "8": "Updated to Node.js v18.",
   "9": "Fixed issue running on devices with older kernels.",
-  "10": "Added sessionless HTTP API auth & automatic dark mode."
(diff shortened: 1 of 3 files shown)


pheiduck:
@suxscribe Thanks for your Patches, the changelog one needs to be cherry-picked.

pheiduck:
@suxscribe Until now, haven't enough time for testing. If you want try out development build and report back if it working or not.

peterlewis:
@suxscribe This PR is looking better by the day. Thank you. Any chance you could make it so that when stats are above buttons, that they are right-aligned to the buttons? Think it would look a little bit neater. Thank you!

suxscribe:
> make it so that when stats are above buttons, that they are right-aligned to the buttons

@peterlewis I think I tried that while working on mobile layout and preferred current alignment. I'll give it another try in a few days

Issue #864, opened 2024-02-21
How does Gravatar support work?


sardaukar:
**Describe the bug**
I created a client named after an email address I know has a Gravatar, but it's still just showing the default avatar icon

**To Reproduce**
1. Add new client
2. Name it after an email address that has a Gravatar image set

**Expected behavior**
The Gravatar image would be shown

**Screenshots**
![image](https://github.com/wg-easy/wg-easy/assets/153407/cb004e2e-f36c-4b06-92aa-7b6c85c64163)

**Desktop (please complete the following information):**
 - OS: [Linux]
 - Browser [Firefox]
 - Version [123]

**Additional context**
N/A

sardaukar:
Opened a PR https://github.com/wg-easy/wg-easy/pull/865 with a proposed fix. I'm not a JS developer, so please go easy on me :D

sardaukar:
Thanks for the merge!

PR #865, opened 2024-02-21
Fix Gravatar support

sardaukar:
After [reporting an issue] with the Gravatar support, I went digging in the code and found the [official docs] on how to get the Gravatar images, and have updated the code to fix it. 

image

Ideally, the name and email would be separate, so multiple users can have the same email for different devices, but this feels like a good start.

[reporting an issue]: https://github.com/wg-easy/wg-easy/issues/864
[official docs]: https://docs.gravatar.com/general/hash/


sardaukar:
Oh, and got the `sha256` JS from https://cdnjs.com/libraries/js-sha256

pheiduck:
@sardaukar Thanks for the Patch I will take a look later

pheiduck (approved):
Let’s wait until @peterlewis approve it as well.

peterlewis (approved):
Approved! Thank you for your PR and I agree with your point regarding multiple devices.

pheiduck:
This PR starts dev freeze for v11.
To proceed on next release.

PR #872, opened 2024-02-22
Fix detailed stats 3

suxscribe:
Aligned detailed stats block to the right on mobile.

There's another option - align everything to the left on mobile. Looks nice to me, what do you think?

<details>
  <summary>Right-aligned</summary>

  right

</details>

<details>
  <summary>Left-aligned</summary>

  left

</details>

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -192,7 +192,7 @@
 
                     <!-- Info -->
                     <div v-if="uiDetailedStats"
-                      class="flex gap-2 items-center shrink-0 text-gray-400 dark:text-neutral-400 text-xs mt-px -ml-4 xxs:ml-0">
+                      class="flex gap-2 items-center shrink-0 text-gray-400 dark:text-neutral-400 text-xs mt-px justify-end">
 
                       <!-- Transfer TX -->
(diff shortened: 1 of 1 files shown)


pheiduck (approved):

peterlewis:
Deffo think right over left. I don't mean to be a pain, sorry, but can I suggest for the speed to also be right aligned? What do you think?

peterlewis:
Just need to give it a test. Will do so when I make it to my desk! :) Thank you!

pheiduck:
> Deffo think right over left! :)

Here as well but left might look better on mobile.

suxscribe:
Keeping right then =)

peterlewis:
Sorry, still awaiting review from me. Haven't forgotten, just have had very little spare time.

peterlewis:
This is looking grand, thank you. I'm just wondering whether `UI_DETAILED_STATS` is the right variable name for this. Yes, they are more detailed, but it's as much about their more visible location as it is the fact that total ingress/egress is made visible. Thoughts?

pheiduck:
@peterlewis `UI_USAGE_STATS` would make more sense than the current name.

PR #880, opened 2024-02-28
refactor!: migrate from express to h3

cany748:
This PR replaces expreess with h3. h3 is a modern, fast, small, written in typescript, cross-platform framework.

To be clearer, this PR makes minimal changes to the:

1. The h3 does not add `charset=utf-8"` to all requests, so `<meta charset="utf-8"/>` is added to index.html. This is present in most modern web pages.
2. Added `return { succcess: true }` to unanswered queries for correct operation.
3. Uses `fromNodeMiddleware` instead of native h3 functions.

We can make the code cleaner:

1. Replace `express-session` with built-in h3 functions. https://h3.unjs.io/examples/handle-session
2. Rewrite the user authentication part using h3's own functions, without using `fromNodeMiddleware`

After adopting this PR we can think about switching to `Nitro`. `Nitro` is a full-featured server toolkit that makes it easy to build a server and all dependencies in a few files, regardless of platform. `Nitro` supports TypeScript and ESM out of the box. Before switching to `Nitro` it will be necessary to update the Eslint config, to an ESM oriented one.

I am happy to answer any questions and can make any changes I suggest.


pheiduck:
@cany748 overall lgtm. I will take a deeper look later. And we should land it in a dedicated branch before `master` for testing if it breaks something or not.

pheiduck (approved):

peterlewis:
I'm not hugely confident as to what the right decision is, with regards to migration away from express. I appreciate there is both this PR as well as #880 and I want to both ensure we're taking it in the best direction, as well as ensuring that we're not wasting your time, too. Would welcome thoughts from all regarding this. Feels like we need to be making an informed choice.

pheiduck:
I still prefer the h3 framework as it looks more maintainable. But let’s make a poll until we make a choice.

pheiduck:
Poll is here: https://github.com/wg-easy/wg-easy/discussions/890
Please vote.

pheiduck:
@cany748 Is there any outstanding development requirement or can we include it in the test branch?

pheiduck:
I update target branch. @cany748 please resolve conflicts.

Issue #859, opened 2024-02-20
Tx/Rx charts not working


DanyaSWorlD:
After one of updates charts have been disappeared, but they are still presented in features list.
![изображение](https://github.com/wg-easy/wg-easy/assets/19343702/0788ed42-0ba1-4b47-8775-8810971b12ce)
So, the question is are they expected to be back or they are removed because of UI bad perfomance and gone forever?

peterlewis:
The absence of charts is not intentional, this appears to be a bug. Thank you for raising!

peterlewis:
Paging @suxscribe as, evident from #872, he is somewhat of a Tailwind guru... ;)

suxscribe:
@peterlewis I've also thought this was intentional. I'll see what i can do
It's not that I like Tailwind this much btw =)

peterlewis:
@pheiduck I'm not sure that https://github.com/wg-easy/wg-easy/pull/873 fixes this, yet? Or am I mistaken?

suxscribe:
I've managed to get charts working. 
Looks quite cluttered, imo. Also the chart moves in steps, I got headache while debugging =)

Attached screenshots of different chart types:
<details>
  <summary>Charts</summary>
  ![2024-02-26_14-46-35](https://github.com/wg-easy/wg-easy/assets/19854626/4e5f75b5-98e4-42c1-948d-5fe7f0d9c1a6)
  ![2024-02-26_14-47-39](https://github.com/wg-easy/wg-easy/assets/19854626/7c466e6e-8c4c-497a-95ba-f2a3a4dc1a6e)
  ![2024-02-26_14-48-50](https://github.com/wg-easy/wg-easy/assets/19854626/d20758c4-fcd4-4247-b450-5739d932fd8f)
  ![2024-02-26_14-51-19](https://github.com/wg-easy/wg-easy/assets/19854626/e35d1589-4d95-4f6b-bb1a-1d5dc15a462a)
</details>

peterlewis:
Ah amazing! Thank you @suxscribe! That is interesting! It may be pertinent for us to re-surface this through an environment variable, such as `UI_CHARTS` and default to `false`.

I note you've included a few examples as well. Do you think it would be a good idea to allow for the changing of styles with `UI_CHARTS_STYLE` or equivilent?

Thank you again, I would really welcome all your thoughts as to the above ponderings.

suxscribe:
@peterlewis Yep, it needs further testing, so variables totally make sense

PR #911, opened 2024-03-13
Fix TX/RX charts. Add UI_CHART_TYPE

suxscribe:
Fixes: #859.

Also adds `UI_CHART_TYPE` option to enable and set chart type:

```
UI_CHART_TYPE=0 # Charts disabled
UI_CHART_TYPE=1 # Line chart
UI_CHART_TYPE=2 # Area chart
UI_CHART_TYPE=3 # Bar chart
```

Needs some testing in terms of performance. On my PC it works fine and only adds about 7-10% CPU and 3-5% GPU load.

<details>
  <summary>Chart screenshots:</summary>

  Bar

  Area

  Line

</details>


pheiduck:
@suxscribe should this be a hotfix for v12?

suxscribe:
@pheiduck Not yet, imo. I'd suggest to add some toggle in the UI to be able to turn charts on/off. What do you think?

pheiduck:
sounds good.

PR #915, opened 2024-03-14
Fix TX/RX charts. Add UI_CHART_TYPE

pheiduck:
From @suxscribe 
Fixes: #859.

Also adds `UI_CHART_TYPE` option to enable and set chart type:

```
UI_CHART_TYPE=0 # Charts disabled
UI_CHART_TYPE=1 # Line chart
UI_CHART_TYPE=2 # Area chart
UI_CHART_TYPE=3 # Bar chart
```

Needs some testing in terms of performance. On my PC it works fine and only adds about 7-10% CPU and 3-5% GPU load.

<details>
  <summary>Chart screenshots:</summary>

  Bar

  Area

  Line

</details>


peterlewis:
Looks good! Just note the one comment left... :)

src/www/js/app.js
```diff
@@ -23,14 +23,33 @@ function bytes(bytes, decimals, kib, maxunit) {
   return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`;
 }
 
+const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
```
peterlewis, line 26: Just checking this section aligns to our use of `media` for automatic dark/light mode?

pheiduck: I guess this is not needed as we defined media auto light dark here: https://github.com/wg-easy/wg-easy/blob/master/src/tailwind.config.js

suxscribe: I didn't quite understand how it works currently :)
The added query is needed to define colors of the charts for light / dark mode.

I can't find where `toggleTheme` method is used, and nothing happens when setting `theme` value in local storage to either `light` or `dark`.

There probably should be a toggle in the UI. The media query could be used to get current browser theme and then switch it with the toggle if needed.

pheiduck:
Auto light/dark mode is working for charts.

suxscribe (approved):
Looks good

suxscribe:
I've made a few updates - added chart and dark/light mode toggle buttons.
Should I commit them in this PR or make a new one?

pheiduck:
@suxscribe sure, if it does not break the auto light/dark mode switch

peterlewis:
@suxscribe before you make a new PR it's worth pointing out that the underlying mechanism and code for light/dark toggle still exists from the merge upstream from fork, we just ripped out the toggle and switched to using `media`. See the following:

https://github.com/pheiduck/wg-easy/commit/de9eebd82703faa4d24bcbedce4e8118897e5ac1

pheiduck:
> @suxscribe before you make a new PR it's worth pointing out that the underlying mechanism and code for light/dark toggle still exists from the merge upstream from fork, we just ripped out the toggle and switched to using `media`. See the following:
> 
> [pheiduck@de9eebd](https://github.com/pheiduck/wg-easy/commit/de9eebd82703faa4d24bcbedce4e8118897e5ac1)

Yes I perfer the new methode anyway.

suxscribe:
@peterlewis Oh okay, I've changed that to `selector` =) . And added ability to switch between auto dark mode and forcing light or dark

PR #930, opened 2024-03-21
refactor!: separate source code

### LINKED ISSUE #928: Source code separation
### What am I suggesting?
I propose to split the source code into two parts: `frontend` and `server`. This shouldn't be a problem, we will need to move a few directories and change a few files. We can keep the existing directory names `src` and `www`, but I would recommend using more contextual names.

### Why?
You need to have a clear separation between the two runtime environments: `Node.js` and `browser`. So that we can use different configurations of Eslint and Typescript, and clearly separate the dependencies.

I'm ready to do a PR.

cany748:
Resolves #928


pheiduck:
@cany748 thanks for your PR, it needs a rebase and some CIs have to updated as well to use the new location of `package.json` and `package-lock.json`
https://github.com/wg-easy/wg-easy/blob/master/.github/workflows/lint.yml
https://github.com/wg-easy/wg-easy/blob/master/.github/workflows/npm-update-bot.yml

cany748:
> @cany748 Is this safe to merge directly into master "without" testing?

yes

peterlewis:
> @cany748 Is this safe to merge directly into master "without" testing?

*Nothing* is safe to merge to master without testing... 😀

pheiduck:
@cany748 please rebase.

Issue #931, opened 2024-03-22
Blank screen on latest `nightly`


peterlewis:
**Describe the bug**
Blank screen on load in latest `nightly`

**To Reproduce**
1. Go to wg-easy 
2. Observe blank screen

**Expected behavior**
wg-easy should display

**Screenshots**
![image](https://github.com/wg-easy/wg-easy/assets/820394/85ddd677-d4d3-4fc1-8fb6-e16e3203b348)

**Smartphone (please complete the following information):**
 - Device: iPhone 14
 - OS: 17.4
 - Browser: Safari

**Additional context**
Haven't done any troubleshooting yet but raising issue before I get started with morning chores.

pheiduck:
Not a problem for me: 
![Bildschirmfoto_2024-03-22_10-17-51](https://github.com/wg-easy/wg-easy/assets/47042125/d963bff7-1c8c-4cc1-8e4c-1d41bc25b58a)

pheiduck:
If you use Nginx Proxy Manager there is an option called "Cache Assets" which should be disabled for wg-easy. Otherwise removing browser data can help.

LoV432:
I am using traefik proxy and the webpage css isn't loading with these errors:
```
The script from “https://wireguard.removed.com/js/vendor/vue.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
The script from “https://wireguard.removed.com/js/vendor/vue-i18n.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
The script from “https://wireguard.removed.com/js/vendor/apexcharts.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
The script from “https://wireguard.removed.com/js/vendor/vue-apexcharts.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
The script from “https://wireguard.removed.com/js/vendor/sha256.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
The script from “https://wireguard.removed.com/js/vendor/timeago.full.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type. wireguard.removed.com
[...]

PR #933, opened 2024-03-23
fix: add Content-Type header for static files

pheiduck:
Merge #932 into `master`
Fixes: #931

### DIFF
--- a/src/lib/Server.js
+++ b/src/lib/Server.js
@@ -225,6 +225,12 @@ module.exports = class Server {
               return;
             }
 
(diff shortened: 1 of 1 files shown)


pheiduck:
@peterlewis if you are around please take your time for testing (no rush) ;)

peterlewis:
No worky.

peterlewis (approved):

pheiduck:
@peterlewis I think the browser cache cleaning is only a thing if Nginx Proxy Manager is used, with Cache Assets enabled. (On my side for eg) I tested it as well and happy to see no debug logs in the browser console with this patch.
Thanks @cany748!

PR #935, opened 2024-03-24
Add support for SVG and include proper Content-Type

Dartegnian:
I noticed #931 appearing on my [wg-easy instance](https://wg.dartegnian.com/) and [fork](https://github.com/Dartegnian/wg-easy-m3). I've seen #933 pushed and I'd like to request that the Content-Type for SVG be included since my fork uses SVGs. It can also help in advance if this project ever includes SVG files.


pheiduck:
Well, I can’t really approve that as we don’t use SVGs… But feel free to apply that on your fork.

PR #939, opened 2024-03-25
fix cidr block calculation issue, fix post down config

utkarsh867:


utkarsh867:
thanks for getting it started @pheiduck! I've been using wg-easy for a bit, and I love it! I've been fiddling around to get this functionality in when I saw this PR. Most of the heavy lifting was done in your PR :)

PR #951, opened 2024-03-31
Fix WG_DEFAULT_ADDRESS if not set

pheiduck:
Fix from @utkarsh867


pheiduck:
legacy WG_DEFAULT_ADDRESS still working with this patch.

PR #957, opened 2024-04-01
Rename screenshot.png to screenshot.svg

pheiduck:
use svg for the README image


pheiduck:
Only Github it self can no render the image under assets

Issue #965, opened 2024-04-05
Error starting container


mgite:
When trying to start latest version it has this error. I have configured both
```
- WG_DEFAULT_ADDRESS=10.8.0.0
- WG_DEFAULT_ADDRESS_RANGE=24
```

But it seems it does not get WG_DEFAULT_ADDRESS_RANGE in the config correctly and it says undefined.

```
$ wg-quick down wg0
$ wg-quick up wg0
Error: Command failed: wg-quick up wg0
[#]
[#] ip link add wg0 type wireguard
[#] wg setconf wg0 /dev/fd/63
[#] ip -4 address add 10.8.0.1/undefined dev wg0
Error: inet prefix is expected rather than "10.8.0.1/undefined".
[#] ip link delete dev wg0
     at genericNodeError (node:internal/errors:984:15)
     at wrappedFn (node:internal/errors:538:14)
     at ChildProcess.exithandler (node:child_process:422:12)
     at ChildProcess.emit (node:events:518:28)
     at maybeClose (node:internal/child_process:1105:16)
     at ChildProcess._handle.onexit (node:internal/child_process:305:5) {
   code: 1,
   killed: false,
   signal: null,
   cmd: 'wg-quick up wg0'
}

```

rudoyeugene:
Nightly build, confirmed

pheiduck:
I hope sometime we get the cidr notation working as expected. For now it is excluded from the next release v13.

utkarsh867:
@mgite @pheiduck I'm unable to reproduce this on my side.

Can you please provide the `wg0.conf` and the `wg0.json` files?

PR #968, opened 2024-04-06
Revert "Fix WG_DEFAULT_ADDRESS if not set"

pheiduck:
Reverts wg-easy/wg-easy#951


pheiduck:
The CIDR Feature is to buggy until now: https://github.com/wg-easy/wg-easy/issues/965

PR #969, opened 2024-04-06
Revert "feat: cidr notation"

pheiduck:


pheiduck:
The CIDR Feature is to buggy until now: https://github.com/wg-easy/wg-easy/issues/965

PR #998, opened 2024-04-18
Update node version : 18 -> 21

comassky:
Update node version, works on my server

Resolve some CVEs


comassky:
Hi @pheiduck , thanks for answers

How can i reproduce build issue on my server ?

docker build command works.

pheiduck:
> How can i reproduce build issue on my server ?

You may set up a virt with armv6/7 architecture to do that.

And we want also not a stable nodejs version as they have a short "life cycles"

PR #1009, opened 2024-04-27
healthcheck is integrated internally 

pheiduck:
Thanks @CosasDePuma


pheiduck:
Some fixups was needed see: https://github.com/wg-easy/wg-easy/commit/c5d328be2a59a4d39584a841b6a261fdf7e22484

PR #1015, opened 2024-04-30
Improved local development and Dockerfile

pheiduck:
Thanks @davide-acanfora


pheiduck:
Perfect it does not break the Docker CI Builds

PR #1024, opened 2024-05-04
dev: bump to node 22 LTS (early build)

pheiduck:
**Don't merge now!**
Still an early test to see if build issue are fixed on newer images.


pheiduck:
Otherwise we could test debian-slim for build step

PR #1036, opened 2024-05-14
fixup: WG_PORT in config.js

### LINKED ISSUE #1032: Unable to connect after upgrade to version 13
With the upgrade to version 13 I cannot connect with my mobile phone to the VPN of my Raspberry Pi5 (Debian bookworm based OS).
In the logs I see the connection but in reality I am not connected, in fact the connection point is not visible.

If I downgrade to version 12 everything starts working properly again, so the portforward in my router is certainly configured correctly.
I await your news

Thank you

### LINKED ISSUE #1045: Wireguard not working in release tagged 13 (latest)
I just updated to version 13, the application works, but Wireguard does not. I rolled back to 12 and everything works again. There are no errors in the logs. I don't know what additional information there is.

pheiduck:
Fixes #1032
Fixes #1045

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ These options can be configured by setting environment variables using `-e KEY="
 | `PASSWORD` | - | `foobar123` | When set, requires a password when logging in to the Web UI. |
 | `WG_HOST` | - | `vpn.myserver.com` | The public hostname of your VPN server. |
 | `WG_DEVICE` | `eth0` | `ens6f0` | Ethernet device the wireguard traffic should be forwarded through. |
-| `WG_PORT` | `51820` | `12345` | The public UDP port of your VPN server. WireGuard will always listen on 51820 inside the Docker container. |
+| `WG_PORT` | `51820` | `12345` | The public UDP port of your VPN server. WireGuard will listen on that (othwise default) inside the Docker container. |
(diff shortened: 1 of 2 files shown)


kaaax0815:
This should also update the README

pheiduck:
@kaaax0815 please take a look at the updated readme, hope it does not confuse...

jhedd:
Might want to also add to the updating section of the README for people updating to release 13.

As a user who came here because things 'just stopped working', when following the recommended update procedure of 

> docker compose up --detach --pull always

there's no indication there's any difference between releases 12 and 13, and a three-word change to the setup options table is going to be easily missed.  Since the only bugs on this issue are closed there's going to be limited visibility for users who are trying to diagnose why their installations simply stopped working.

pheiduck:
> Might want to also add to the updating section of the README for people updating to release 13.
> 
> As a user who came here because things 'just stopped working', when following the recommended update procedure of
> 
> > docker compose up --detach --pull always
> 
> there's no indication there's any difference between releases 12 and 13, and a three-word change to the setup options table is going to be easily missed. Since the only bugs on this issue are closed there's going to be limited visibility for users who are trying to diagnose why their installations simply stopped working.

Please make PR to include your recommendation.

Issue #1067, opened 2024-05-24
Security vulnerability found in the Web UI


davide-acanfora:
Hi, I just found a vulnerability involving the Web UI when it is publicly exposed to a malicious attacker. Since the flaw allows access to any file inside the container (even files containing client keys) without logging in, I don't know if this is the right place to talk about it. Should I contact someone or can I describe it here?

pheiduck:
Please contact @WeeJeWel. Are you able to make fix for that?

pheiduck:
We have a signal channel maybe we could invite you.

peterlewis:
Unlocked conversation as otherwise @davide-acanfora will not be able to reply, I don't think...

@davide-acanfora, do you have Signal?

davide-acanfora:
No problem, I had planned to email @WeeJeWel, but then I got carried away with fixing the code.

@pheiduck yes https://signal.me/#eu/vY_YuF5sAp2wKLBCxAlIAZe6eBt8jvjBcnbn_vsyoxtIs4wPrtZE54uQ84okIVgH

peterlewis:
Thanks, Davide. I've invited you to the Signal channel.

peterlewis:
Thank you, @davide-acanfora for your hard work on resolving this so quickly!

PR #1071, opened 2024-05-27
Security vulnerability patched and minor improvements

davide-acanfora:
In reference to #1067, this PR fixes the **directory traversal** (**path traversal**) vulnerability found in the last released version and introduced with bad path handling in the h3 framework's `serveStatic` directive. Now the requested resource paths are handled correctly with the `safePathJoin` function. It takes the base path of the static assets and the path to the requested resource and checks that their union is indeed still a valid **subpath**.

In addition, some minor improvements have been implemented:

- replaced the `uuid` package with the built-in Node `crypto` module. Since its only task was to generate **UUIDv4** for new clients, we can use the `randomUUID` function instead;
- removed the `bcryptjs` package. The only reason it was being used was to implement an authorization mechanism by checking that in the `Authorization` header of requests (beginning with `/api/`) there was the correct hash of the dashboard password. From what I've seen wg-easy handles authentication via **session cookies** instead, plus I can't find any part of the client code where that header is set, so I guess both the checking part and the module can be removed;
- stop printing the version of wg-easy in the browser console. I think there is no need to let anyone who visits the page without logging in know about it;

### DIFF
--- a/src/lib/Server.js
+++ b/src/lib/Server.js
@@ -3,7 +3,7 @@
 const crypto = require('node:crypto');
 const { createServer } = require('node:http');
 const { stat, readFile } = require('node:fs/promises');
-const { join } = require('node:path');
+const { resolve, sep } = require('node:path');
 
 const expressSession = require('express-session');
 const debug = require('debug')('Server');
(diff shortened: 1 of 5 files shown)


pheiduck (approved):

pheiduck:
I pulled in the latest commit. only npm updates...

peterlewis:
This is excellent work. Thank you so much, @davide-acanfora.

pheiduck:
Ya, Thanks a lot to take your time to do that kind of security patch @davide-acanfora!

pheiduck:
@peterlewis, could you approve this as well? So we can merge this asap. :)

peterlewis (approved):

pheiduck:
I will trigger then a `dev` / `nightly` build to test, if all seams fine. Merged to `prod`.

peterlewis (commented):
All looks good to me. I note switch from using `uuid` to `crypto` library.

peterlewis (approved):
Looks good to me, I note switch from using `uuid` to using `crypto`.

davide-acanfora:
> Ya, Thanks a lot to take your time to do that kind of security patch @davide-acanfora!

I don't consider myself a security expert, but the solution I proposed should be sufficient. I would be happy if someone with more experience than me could prove otherwise

PR #1085, opened 2024-06-04
Inform Podman users about cap-add=NET_RAW

iguanajuice:
First off, great piece of software! But as a Podman user, it took me longer than I'd like to figure out what was going wrong.

There is only one mention of Podman on the issue tracker, and the "solution" is a broken link. It is mentioned on the wiki, however; but I feel most won't check there, or at least, not at first.


pheiduck:
Oops I added **Readme** to the commit message but it is **docker-compose.yml**

Issue #1083, opened 2024-06-03
[FEATURE PROPOSAL] handle /16 & /8 Subnets


gganster:
Hi,

For IOT purpose with lot of rasberry pi (>255) and VNC needs. Ive modified wg-easy for /16 mask and unlock habilities to have more than 65 000 clients. 

Are you interested by a PR ?

No more configuration needed and retro compatible. It use your existing WG_DEFAULT_ADDRESS env variable to define the mask.

ex: 
4.2.0.x  ->  /24
4.2.x.x  ->  /16

Happy coding.

pheiduck:
Are you interested by a PR ?
Yep 😄

gganster:
Hi,

PR created #1087

bmarinov:
Hey, I just hit this issue and sadly this wouldn't be enough to cover my case. I am on a `/20` network and just realized, that wg-easy generates configuration for `/24`. Can we also add a separate, optional configuration value for the subnet? Out of scope here, but definitely needed IMO.

Many valid network configurations cannot be inferred, especially once you start adding VLANs into the mix and separate in smaller networks. Im planning to split my /20 in two /21 subnets for example.

kaaax0815:
Implemented with #1347

PR #1087, opened 2024-06-04
[FEATURE PROPOSAL] handle /16 & /8 Subnets

gganster:
Hi,

For IOT purpose with lot of rasberry pi (>255) and VNC needs. Ive modified wg-easy for /16 mask and unlock habilities to have more than 65 000 clients.

No more configuration needed and retro compatible. It use your existing WG_DEFAULT_ADDRESS env variable to define the mask.

ex: 4.2.0.x -> /24, 4.2.x.x -> /16

Documentation was added in readme.

Ive used this little script to create a lot of clients and test by hand the new "compute next ip algorithm".

```shell
#!/bin/bash

n_create=1000 # number of iteration
wg_adress="http://192.168.10.129:51821" #ip and port of wireguard http server
wg_password="yourpasswordhere"

(curl -c cookies.txt -H "Content-Type: application/json" -d "{\"password\":\"$wg_password\"}" -X POST $wg_adress/api/session --silent) > /dev/null
connect_sid=$(awk '/connect.sid/ {print $NF}' cookies.txt)

echo "connect.sid: $connect_sid"

for i in $(eval echo "{1..$n_create}")
do
  hostname=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
  (curl -H "Cookie: connect.sid=$connect_sid" -H "Content-Type: application/json" -d "{\"name\":\"$hostname\"}" -X POST $wg_adress/api/wireguard/client --silent) > /dev/null
  res=$(curl -H "Cookie: connect.sid=$connect_sid" -H "Content-Type: application/json" -X GET $wg_adress/api/wireguard/client --silent)
  clientObject=$(echo $res | jq -r ".[] | select(.name == \"$hostname\")")
  echo "$clientObject"
done
``` 

reference #1083


pheiduck:
Please fix lint errors. 

```bash
npm run lint —- —-fix
```

peterlewis:
Thank you for this! Needs testing.

kaaax0815:
I think #1087 would be easier for users to use, but #1003 is more technical correct, as it would support more uncommon cidr ranges.
Therefore I would prefer #1003

pheiduck:
@gganster please rebase.

BloodyHellcat:
Just hitting the topic with the following from the issue section:
https://github.com/wg-easy/wg-easy/issues/1083#issuecomment-2230862087
As he mentioned, the interface is also setup with /24.
This is what currently is blocking my setup aswell as of the following error
`wg-easy  | [#] ip link add wg0 type wireguard`
`wg-easy  | [#] wg setconf wg0 /dev/fd/63`
`wg-easy  | [#] ip -4 address add 192.169.1.x/24 dev wg0`
`wg-easy  | Error: inet prefix is expected rather than "192.169.1.x/24".`
`wg-easy  | [#] ip link delete dev wg0`

My setup would need /16 to work corectly.

If wished, happy to test aswel.

PR #1090, opened 2024-06-07
config: Add support for custom client port configuration 

pheiduck:
By @adriy-be

This commit introduces the ability to specify a custom port for the client configuration. This feature is particularly useful when the WireGuard server is behind a port forwarding setup, allowing clients to connect using the correct port number.

With this change, users can now define the desired client port in the configuration file with the env variable WG_CONFIG_PORT, ensuring seamless connectivity even in scenarios where the client's listening port differs from the standard port.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       # - PASSWORD=foobar123
       # - PORT=51821
       # - WG_PORT=51820
+      # - WG_CONFIG_PORT=92820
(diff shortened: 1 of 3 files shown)


pheiduck:
@adriy-be, please test with `development` tag in docker.

adriy-be:
I have tried with https://github.com/adriy-be/homeassistant-addons-jdeath/tree/main/wgeasy and that working fine. Thanks

PR #1101, opened 2024-06-13
Dockerfile: use native nodejs 20 LTS for build as well

pheiduck:
Thanks @pabra


pheiduck:
$BUILDPLATFORM is using Host arch **amd64** this not what we want.

pheiduck:
Sill take a long time to build...

PR #1145, opened 2024-07-04
translations: Update Traditional Chinese translations

bluehomewu:
Hello, all.
The word usage of the translation strings in the past are in the style of China rather than Taiwan.
I'm a Taiwanese, who works on Android and Translation projects.


pheiduck (approved):
Haven't experience I this kind of languages, but lgtm!
Thank you @bluehomewu!

PR #1162, opened 2024-07-10
Revert "import & export configuration"

pheiduck:
User interface options aren't visable...
@tetuaoro please fix


pheiduck:
Maybe I merge it to fast...

PR #1164, opened 2024-07-10
start fix pr 1161

tetuaoro:
[cannot login](https://github.com/wg-easy/wg-easy/pull/1161#issuecomment-2220234346)


tetuaoro:
ah, yeah my bad

PR #1182, opened 2024-07-14
Automate IP Retrieval for Docker Run Command

idevanshu:
Automatically retrieve and use the server's public IP address in the Docker run command

- Added a step to fetch the server's public IP address using curl dynamically.
- Updated the Docker run command to use the retrieved IP address for the WG_HOST environment variable.


pheiduck:
This is only needed if running on a VPS right?
Some users uses dyndns if running at home.

PR #1201, opened 2024-07-20
Fix #1111

### LINKED ISSUE #1111: Line Chart makes UI look weird
**Describe the bug**
Using chart type 1 makes ui look weird

**To Reproduce**
1. Set chart type 1
2. log in

**Expected behavior**
should look like the other types

**Screenshots**
firefox_oja0rxVRBS

**Additional context**
There are additional lines between each client which should not happen

suxscribe:
Move chart bottom out of container, so it won't show up when rx/tx rate is 0

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -15,6 +15,9 @@
   [v-cloak] {
     display: none;
   }
+  .line-chart .apexcharts-svg{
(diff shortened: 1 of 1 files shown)


pheiduck:
Mercy!

PR #1108, opened 2024-06-16
ipv6 support v2

pheiduck:
Reworked by @kaaax0815
Please test with `development` tag

PR #926, opened 2024-03-19
Upgrade to Vue3

pheiduck:
closed #754 by mistake
anyway needs rebase @oplexz

PR #1202, opened 2024-07-21
update ipv6 support v2

kaaax0815:
updates #1108

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -39,22 +39,35 @@ if (!process.env.WG_POST_UP) {
   iptables -A FORWARD -o wg0 -j ACCEPT;`;
 
   if (modules.includes('ip6table_nat')) {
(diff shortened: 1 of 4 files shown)


kaaax0815:
Most of the functions were already added in my original pr but were later removed likely by a force push or bad merge

kaaax0815:
typescript and #926 would have catched the regression which shouldn't have happened

trmdi:
Does this PR keep the existing conf?

trmdi:
Also, if some people want to disable ipv6, does this PR handle that case?

kaaax0815:
> Does this PR keep the existing conf?

not 100% sure but i think not.

trmdi:
> > Does this PR keep the existing conf?
> 
> not 100% sure but i think not.

Then the old question, since the new conf is just the old one plus new ipv6 addresses, why can't we just add them instead of resetting everything?

Issue #1197, opened 2024-07-18
Cannot find module '/app/wgpw'


SonnieP:
Per the instructions I tried to create a bcrpyt hash password... trying that command gives me the following error:

[root@host sonnie]# docker run ghcr.io/wg-easy/wg-easy wgpw (I entered my password here)
docker run ghcr.io/wg-easy/wg-easy wgpw (my password was shown here)
node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module '/app/wgpw'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_ma                    in:174:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.15.1
[root@host sonnie]#

Any ideas?

eandersons:
That is because you are using image tag that does not have (yet) this functionality. Try `nightly`.

SonnieP:
I'm not following... where do I insert "nightly," and how does that fix the module not found issue? Thanks!

eandersons:
`docker run ghcr.io/wg-easy/wg-easy:nigthly wgpw`

If tag is not specified, `latest` is implied, and currently the `latest` tag does not yet has the functionality you are trying to use.

eandersons:
Did you run container with the image of same tag?
Did you run container with `docker run` or with Compose file?

I suspect that the culprit is that the password hash contains `$`. See #1181 or #1191 depending on how are you running WireGuard Easy container for a solution.

Confusion comes from the fact that the branch `master` is not the production branch, therefore information that can be found on https://github.com/wg-easy/wg-easy might not reflect what is available on production images.

SonnieP:
I ran the command: docker compose up -d && docker compose logs -f

I have the single quote around my hash password and I have cleared cache on my browser. I still get unauthorized. 

I think it's about time I give up on this... not worth two days of my time. I suppose I'll install Wireguard thru my Omada Controller. I thought maybe this would be easier and better, but it has wasted a vast amount of my time and is anything by easy. I never seen a program have so many issues. Truly remarkable that anyone is using it.

mcmimik:
@eandersons There is a typo in your comment (`h` should be before the `t`).

[...]

Issue #1216, opened 2024-07-26
add database


offizium-berndstorath:
**Describe the bug**
Many features are requested that would need to store data somewhere persistent. should be atomic and support multiple connections. so multiple admins wont interfere with one another

a json file is imo not the best storage solution. even a simple (type-safe) wrapper around such file would greatly improve dx.

i looked at some solutions but cant choose:
https://github.com/appy-one/acebase (unmaintained?)
https://github.com/Level/level
https://github.com/kriasoft/node-sqlite (unmaintained?)
https://github.com/typicode/lowdb

kaaax0815:
Implemented with https://github.com/wg-easy/wg-easy/pull/1330

PR #937, opened 2024-03-24
Updates to "Upgrade to Vue3" #926

suxscribe:
Rebase and update of @oplexz Vue3 PR #926 

What have been made:
- Rebased the code
- Split App.vue into multiple components
- Added Pinia store
- Integrated i18n
- Added vue-timeago3
- Updated all layouts and features up to the current state of the master branch

Everything should be functional. Anyway further refactoring and improvements are needed

PR #1228, opened 2024-07-30
feat: UI configure password on first startup + database

tetuaoro:
## Description

Removed the (required) `PASSWORD_HASH` environment variable from the docker config. Use the Web UI for handling password and setting up the username and password at first start.

## Motivation and Context

This change simplifies the deployment process and improves user experience. By handling password hashing through the UI, we eliminate the need for setting environment variables, which can be cumbersome and error-prone, and resolve a lot of related issues (#1197, #1235 ...).

## How has this been tested?

Tested by running the application, opening the `Settings` menu and setting your username and password.

### First setup
image

### Settings menu
image

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


tetuaoro:
Oh you tried a lot, seems arm (v7 is not supported)[https://github.com/TryGhost/node-sqlite3?tab=readme-ov-file#prebuilt-binaries]. My bad

pheiduck:
Yep armv6 too, will revert my changes

tetuaoro:
ok, switch to simple json file hahaha

pheiduck:
We have to find a database to support all our architectures or move on drop these but this is uncool 🙃

pheiduck:
alpine sqlite3 support all architectures.
https://git.alpinelinux.org/aports/tree/main/sqlite/APKBUILD?id=898038f2bb7eae5ebd315cde97cac25c3e66a757
I don't know if these are usable in nodejs.

pheiduck:
Please try this flag: --build-from-source=sqlite3 otherwise everything is build from source. If it possible to reduce the buildtime

pheiduck:
> The password is required and the default password is 'admin'.

This will be a security hole, the Password should be choose by the user.

tetuaoro:
I will use json file until we find a great and supportd database provider.

tetuaoro:
> using a supported database would solve #1216 and would greatly increase dx for new features. maybe adding a db would be great for a separate pr? then things like all the env vars could be configured in the ui
> 
> maybe try https://orm.drizzle.team/docs/get-started-sqlite#better-sqlite3 (drizzle would add good dx and safety) or [`better-sqlite3`](https://www.npmjs.com/package/better-sqlite3) alone. or use drizzle with a e.g. postgres db for better performance
> 
> for the security considerations. other projects generate a random password on first start and store it in a file

Hi & thank you, I'll try next week.

tetuaoro:
Can you refine your idea about the password ? How should we proceed ?

tetuaoro:
> off-topic: really like that you created a database interface to support multiple databases in the future

Yes, I use an abstract class to do this. I read your post #1216. I'll try first if a service in the container would do, like postgres or mysql. I'm talking about weight.

tetuaoro:
> > I'll try first if a service in the container would do, like postgres or mysql. I'm talking about weight.
> 
> no need for a db inside the wg-easy container. if you want to use postgres over sqlite then spin up a new container.

Have you tried `lowdb` (ref #1216) ? It use json file.

pheiduck:
`lowdb` seems more up-to-date

tetuaoro:
> very good work @tetuaoro as the dashboard will/should reuse parts of other ui. maybe #937 is needed to split ui into components

yes, I would like to write components and esm modules.

tetuaoro:
Done.

`PASSWORD_HASH` has been completely deleted. You configure a username and password on first startup (if database does not exist).

tetuaoro:
Waiting for #1244

kaaax0815:
@tetuaoro you can recreate this pr against `feat-ui`

Issue #1234, opened 2024-08-01
Docker-compose dev not working


offizium-berndstorath:
**Describe the bug**
`docker compose -f docker-compose.dev.yml up` results in error on first run. as `wg-easy` image is need to run

**To Reproduce**
1. Clone
2. Run

**Expected behavior**
you can ofc run `docker build` but that doesnt make much sense
just being able to run `docker compose -f docker-compose.dev.yml up --build` makes most sense. as it also builds the dockerfile which is needed when e.g. switching branches with different features. also ports are not exposed and permissions are missing.
a docker compose like this makes most sense

```yaml
services:
  wg-easy:
    build:
      dockerfile: ./Dockerfile
    command: npm run serve
    volumes:
      - ./src/:/app/
      # - ./data/:/etc/wireguard
    ports:
      - "51820:51820/udp"
      - "51821:51821/tcp"
    cap_add:
      - NET_ADMIN
      - SYS_MODULE
    environment:
      # - PASSWORD=p
      - WG_HOST=192.168.1.233

```

PR #1236, opened 2024-08-01
fixup: docker-compose.dev.yml

pheiduck:
Improving compose file for dev.
Some variables are missing which lead to an error. 
Fixes: #1234


pheiduck:
@offizium-berndstorath please take a look.

PR #1246, opened 2024-08-05
Changes made in readme.md about changing WEBUI_HOST parameter 

Inforegister:
Changes made in readme.md about changing WEBUI_HOST parameter to UI_HOST and changed description of WG_HOST


pheiduck:
hm, this make no sense due the variable `UI_HOST` does not exist...

PR #1249, opened 2024-08-07
Add autocomplete attribute to password input

ibrahimcetin:
Safari suggests strong password on login page. This PR should fix this problem.

strong password suggestion


pheiduck (approved):
@ibrahimcetin Thank you!

PR #1250, opened 2024-08-07
Split components

tetuaoro:
I introduced `pages` and `components` directories. Why pages ? Because I'll use them in my next PR. I don't know if you think the same way for components. This is just an introduction, and I need feedback before continuing with the splitting.


tetuaoro:
Ah, I can't see the `feat-nuxt` branch.

PR #1270, opened 2024-08-12
fix(Doc): fix escaping issue for PASSWORD_HASH in docker-compose.yml

cyicz123:
Hello! I noticed repeated changes to docker-compose.yml in previous commits. When applying the single-quote pattern, I encountered the following warning:

`WARN[0000] The "hBCoykrB95WSzuV4fafBzOHWKu9sbyVa34GJr8VV5R" variable is not set. Defaulting to a blank string.`

Additionally, the `PASSWORD_HASH` variable value was incorrect when viewed inside the container.

# Solution:
Remove the single quotes and insert an extra `$` before each existing `$`. After this modification, the program works as expected.

# Possible cause:
There might be inconsistencies in Docker Compose versions between us, leading to different behavior in handling escape characters.

My environment information:
```bash
docker --version
Docker version 27.1.1, build 6312585
docker compose version
Docker Compose version v2.29.1
```

The changes in this PR work correctly with my Docker Compose version. Please review and confirm if it also works properly with other versions.
#1261


pheiduck:
@cyicz123 Thank you for the doc update. For me the current behavior is working due I have my variables in an `.env` file. Could you please mention this to in the doc?

PR #1274, opened 2024-08-12
add better issue template

offizium-berndstorath:
This permits new users only to a specific format of issues and questions have to be asked using discussions.

proof read requested


pheiduck (approved):
@offizium-berndstorath Thank you!

PR #1276, opened 2024-08-12
[FEATURE PROPOSAL] Remember me

spcfox:
## Description
Added a "Remember me" button for long sessions.

Currently session cookies live until the browser is closed. This PR adds a setting that allows you to set the cookie max age. If it is set to 0, then everything will be as before. If the user has set custom value, then upon login a "Remember me" checkbox will appear.

**Possible addition**:
Also, maybe some users want to always use short sessions (for example, a few minutes) to increase security. In this case, they would benefit from the ability to change MAX_AGE, but disable the "Remember me" button. If you think this would be useful, I can implement it.

## Motivation and Context
Some users do not like to log in after each browser restart. They would like to always remain logged in on their devices.

## How has this been tested?
Tested manually, running with different MAX_AGE values ​​and checking that the behavior matches the expected one. My PR minimally affects the existing code.

image

## Types of changes

## Checklist:


pheiduck:
@spcfox thank you for this PR.

pheiduck:
@spcfox please rebase

pheiduck (approved):

Issue #1287, opened 2024-08-16
[Feat]: Create User with Certificate Expiration Date


eliasmainetti:
### Description

I would like to have the option to have the Certificate validity at the time of user creation. With this, we could send the key to the user with a validity of 1 month, for example. This is good for VPN use cases with specific periods.

### Suggested Solution

When creating the certificate, add the certificate validity period.

In the WebUI Interface, display the validity period of the client certificate.

PR #1296, opened 2024-08-18
Feat expiration date

AKEB:
New feature, allows you to specify the VPN validity period for the client

## Description

When environment WG_ENABLE_EXPIRES_TIME is enabled, the certificate expiration date is displayed in the web interface. Once a minute a check for an expired certificate is performed and the client with an expired certificate is disabled. 

## Motivation and Context

https://github.com/wg-easy/wg-easy/issues/1287

## How has this been tested?

Tested this on my computer

Screenshot at Aug 18 22-37-50
Screenshot at Aug 18 22-37-20
Screenshot at Aug 18 22-36-21

## Types of changes

## Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# The JSON files contain newlines inconsistently
+[*.json]
+insert_final_newline = ignore
+
+# Minified JavaScript files shouldn't be changed
+[**.min.js]
+indent_style = ignore
+insert_final_newline = ignore
+
+[*.md]
+trim_trailing_whitespace = false
+
--- a/src/lib/Server.js
+++ b/src/lib/Server.js
@@ -35,6 +35,7 @@ const {
   UI_CHART_TYPE,
   UI_SHOW_LINKS,
   UI_ENABLE_SORT_CLIENTS,
+  WG_ENABLE_EXPIRES_TIME,
 } = require('../config');
 
 const requiresPassword = !!PASSWORD_HASH;
@@ -57,6 +58,11 @@ const isPasswordValid = (password) => {
   }
 
   return false;
+};
+
+const cronJobEveryMinute = async () => {
+  await WireGuard.cronJobEveryMinute();
+  setTimeout(cronJobEveryMinute, 60 * 1000);
 };
 
 module.exports = class Server {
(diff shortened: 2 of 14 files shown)


pheiduck (approved):

.editorconfig
everythingallthetime, line 1: Why is this in here?

src/lib/Server.js
```diff
@@ -59,6 +60,11 @@ const isPasswordValid = (password) => {
   return false;
 };
 
+const cronJobEveryMinute = async () => {
```
everythingallthetime, line 63: Why is this not using [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/setInterval)?

Why is the interval hardcoded to run every single minute of every day if the expiry time is hardcoded to 23:59.50 anyway?

AKEB: It doesn't matter whether to use setInterval or setTimeout. There is no difference between them in this implementation. If it is so necessary, we can change it to setInterval


If you run the function once a day, there will be a lag between the blocking time and the actual time when the function was run. And it can be another 24 hours if docker was started at 23:59:49.
And now there is a lag of only 1 minute. 

You can probably make it 10 minutes. Then the lag will be 10 minutes from the time when you need to lock to the time of actual locking.

.editorconfig
everythingallthetime: I don't think it's good practice to push editor configs into the project in a feature PR, though.

Re. `setInterval()`:
>There is no difference between them in this implementation.

There absolutely is, `setTimeout()` requires this superfluous function call: https://github.com/wg-easy/wg-easy/blob/352a022f303528be1edcecbbac97a4af2d6bd196/src/lib/Server.js#L365

Re. expiry interval: Wouldnt' it be best  to rename `cronJobEveryMinute()`  to something more meaningful and more general and parameterize it?  Something like  `checkExpiredClients(interval)` and make `interval` user-configurable with a sane default of 10 or 15 minutes.

Please know I'm not trying to shit on your work here (I like this feature, and PR has already been merged after all), just trying to increase maintainability a bit.

AKEB: Thanks for the comment. I think it can be unfixed now. If someone needs to increase the time, or change it for some reason, then we'll fix it then

IljaBekirov:
I deployed the docker image of WG_ENABLE_EXPIRES_TIME , but this functionality did not appear. 
I used the command:
run docker -d \
 --name=wg-easy \
 -e LANG=ru \
 -e WG_HOST=IP \
-e PASSWORD_HASH=transfer \
 -e WG_ENABLE_EXPIRES_TIME=true \
 -e UI_CHART_TYPE=2 \
-e WG_ENABLE_ONE_TIME_LINKS=true \
 -e UI_ENABLE_SORT_CLIENTS=true \
 -e UI_TRAFFIC_STATS=true \
 -e PORT=51821 \
 -e WG_PORT=51820 \
 -v ~/.wg-easy:/etc/wireguard \
 -page 51820:51820/udp \
 -page 51821:51821/tcp \
 --cap-add=NETWORK ADMINISTRATOR \
 --cap-add=SYSTEM MODULE \
 --sysctl="net.ipv4.conf.all.src_valid_mark=1" \
--sysctl="net.ipv4.ip_forward=1" \
--restart, unless -stopped \
ghcr.io/wg-easy/wg-easy

изображение

AKEB:
Please use ghcr.io/wg-easy/wg-easy:nightly or waiting release version 2

IljaBekirov:
Thanks

PR #1320, opened 2024-08-26
[feature] amneziawg support

CodingJediKnight:
Add support of amneziawg config [ref](https://github.com/amnezia-vpn)

## Description
Add config envs to client config file generation

## Motivation and Context
There is a DPI ban in some contries that blocks incomming packets via wireguard. This is one of the working clients, that helps avoid this. It sends junk random packets to prevent blocking. This packet generation controls via this params in client config file.

## How has this been tested?
via local env

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 /config
 /wg0.conf
 /wg0.json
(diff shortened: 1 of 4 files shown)


pheiduck:
Gets not approved, because:

- Edit Template.
- We have no active co op with the sub Project.

PR #1321, opened 2024-08-27
refactor: don't use the layouts folder

cany748:
## Description
I moved the Header and Footer to the components/Layout folder, in order to use them as normal components.

## Motivation and Context
This change is necessary to prevent misuse of Nuxt tools.
Documentation link: https://nuxt.com/docs/guide/directory-structure/layouts
Usage examples from the Nuxt team: https://github.com/nuxt-ui-pro/saas/tree/main/app/layouts
https://github.com/nuxt-ui-pro/saas/tree/main/app/components

## How has this been tested?
This has been tested.

## Types of changes

## Checklist:


cany748:
why do you even need to use layouts? what's the point? you want this code to execute every time without doing anything useful? https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/components/nuxt-layout.ts

kaaax0815:
There is a reason why the Layouts feature exists. Then it should be used. Optimizing for miniscule performance improvements doesn't make sense, if it adds logical separation

cany748:
I didn't say that the Layouts feature is not needed. what I'm trying to say is that the Layouts feature is now not needed in this application. you should use component names for logical separation.

tetuaoro:
I'll introduce multiple routes like `/` (index), `/dashboard` and `/dashboard/settings`. Maybe an `AppHeader` should be a component and `Footer` should remain a layout... imo.

kaaax0815:
I'll close this pr for now. Will take it into consideration. Feel free to reopen if the new ui is more polished and all ideas are clear

PR #1330, opened 2024-08-29
add database

tetuaoro:
I add `in-memory` and `lowdb` database providers and move `PASSWORD_HASH` from `env` to the user interface *(not `/setting` page)*. If no `users` exist in the database, you will be redirected to the first `/setup` page.

Related to #1216, task #1333 and this from the previous PR #1228.

### Changes

- [x] add `in-memory` & `lodb` database
- [x] first setup page
- [x] create new account
- [x] login with username and password.


src/server/api/session.post.ts
```diff
@@ -14,7 +14,17 @@ export default defineEventHandler(async (event) => {
...
   }
-  if (!isPasswordValid(password, PASSWORD_HASH)) {
+
+  const users = await Database.getUsers();
+  const user = users.find((user) => user.username == username);
+  if (!user)
+    throw createError({
+      statusCode: 400,
+      statusMessage: 'User with username does not exist',
+    });
```
kaaax0815, line 24: this should be made more secure. this would allow bad actors to probe correct usernames.
also timing should be considered

src/server/api/lang.get.ts
```diff
@@ -1,4 +1,4 @@
-export default defineEventHandler((event) => {
+export default defineEventHandler(async (event) => {
   setHeader(event, 'Content-Type', 'application/json');
-  return LANG;
+  return await Database.getLang();
```
kaaax0815, line 3: await is not needed. as returning a Promise is fine

src/server/api/account/new.post.ts
```diff
@@ -0,0 +1,22 @@
+import { DatabaseError } from '~/ports/database';
+
+type Request = { username: string; password: string };
+
+export default defineEventHandler(async (event) => {
+  setHeader(event, 'Content-Type', 'application/json');
+  try {
+    // TODO use zod
```
kaaax0815, line 8: should be done. before merging. also please use `TODO: message_here`

src/ports/user/model.ts
```diff
@@ -0,0 +1,28 @@
+import type { Address, ID, Key, HashPassword, String, Boolean } from '../types';
+
+export enum ROLE {
+  /* Full permissions to any resources (app, database...) */
+  ADMIN = 'ADMIN',
+  /* Grants write and read permissions on their own resources and `CLIENT` resources without `ADMIN` permissions */
+  EDITOR = 'EDITOR',
+  /* Grants write and read permissions on their own resources  */
+  CLIENT = 'CLIENT',
+}
```
kaaax0815, line 10: personally think enums are not very good in typescript [^1]

[^1]: https://graphqleditor.com/blog/enums-are-still-bad/

src/ports/types.ts
```diff
@@ -0,0 +1,56 @@
...
+export enum Lang {
+  /* english */
+  EN = 'en',
+  /* french */
+  FR = 'fr',
+}
+export type String = string;
+export type Number = number;
+export type ID = String | Number;
+export type Boolean = boolean;
```
kaaax0815, line 11: Id is a uuid therefore a string.
Also I don't think there is a reason for creating a type alias for basic types

src/ports/types.ts
```diff
@@ -0,0 +1,56 @@
...
+export type Version = String;
+export type SessionTimeOut = Number;
+export type Port = Number;
+export type Address = String;
+export type HashPassword = String;
+export type Command = String;
+export type Key = String;
+export type IpTables = {
+  wgPreUp: Command;
```
kaaax0815, line 20: `wgPreUp: string` is totally fine

src/ports/types.ts
```diff
@@ -0,0 +1,56 @@
...
+  address: Address;
+};
+export type WGConfig = {
+  mtu: Number;
+  persistentKeepalive: Number;
+  rangeAddress: Address;
+  defaultDns: Array<Address>;
+  allowedIps: Array<Address>;
+};
+export enum ChartType {
```
kaaax0815, line 37: see other comment about enums

src/ports/database.ts
```diff
@@ -0,0 +1,53 @@
+import type SystemRepository from './system/interface';
+import type UserRepository from './user/interface';
+import type { Identity, Undefined, Lang, String } from './types';
+import type { User } from './user/model';
+import type { System } from './system/model';
+
```
kaaax0815, line 6: question about the folder structure.
what does `ports` mean in this context?
maybe everything database related should be moved into a single folder to not "pollute" the folder structure

src/ports/database.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Abstract class for database operations.
+ * Provides methods to connect, disconnect, and interact with system and user data.
+ *
+ * *Note : Throw with `DatabaseError` to ensure API handling errors.*
```
kaaax0815, line 11: use jsdoc for that

src/pages/setup.vue
```diff
@@ -0,0 +1,60 @@
...
+        </div>
+        <button type="submit">Save</button>
+      </form>
+    </div>
+  </main>
+</template>
+
+<script setup lang="ts">
+const username = ref<string>();
+const password = ref<string>();
```
kaaax0815, line 39: its common practive in this repo to use null as a initial value

src/pages/login.vue
```diff
@@ -77,19 +86,25 @@
 <script setup lang="ts">
 const authenticating = ref(false);
 const remember = ref(false);
+const username = ref<string>();
```
kaaax0815, line 89: use null as initial value

src/ports/types.ts
```diff
@@ -0,0 +1,56 @@
...
+export type Prometheus = {
+  enabled: Boolean;
+  password?: HashPassword | Undefined;
+};
+/**
+ * `id` of T or T.
+ *
+ * @template T - The specific type that can be used in place of id String.
+ */
+export type Identity<T> = ID | T;
```
kaaax0815, line 56: Why not use string directly?

src/ports/database.ts
```diff
@@ -0,0 +1,53 @@
...
+    _username: String,
+    _password: String
+  ): Promise<void>;
+  abstract saveUser(_user: User): Promise<void>;
+  abstract deleteUser(_id: Identity<User>): Promise<void>;
+}
+
+export class DatabaseError extends Error {
+  static readonly ERROR_PASSWORD_REQ =
+    'Password does not meet the strength requirements. It must be at least 12 characters long, with at least one uppercase letter, one lowercase letter, one number, and one special character.';
```
kaaax0815, line 42: this should probably done using zod. or as a key for i18n

src/ports/database.ts
```diff
@@ -0,0 +1,53 @@
...
+  abstract getSystem(): Promise<System | Undefined>;
+  abstract getLang(): Promise<Lang>;
+
+  abstract getUsers(): Promise<Array<User>>;
+  abstract getUser(_id: Identity<User>): Promise<User | Undefined>;
+  abstract newUserWithPassword(
+    _username: String,
+    _password: String
+  ): Promise<void>;
+  abstract saveUser(_user: User): Promise<void>;
```
kaaax0815, line 36: avoid using underscore for variables

src/adapters/database/inmemory.ts
```diff
@@ -0,0 +1,159 @@
...
+import type { SessionConfig } from 'h3';
+import type { System } from '~/ports/system/model';
+import type { User } from '~/ports/user/model';
+import type { Identity, String } from '~/ports/types';
+import {
+  hashPasswordWithBcrypt,
+  isPasswordStrong,
+} from '~/server/utils/password';
+
+const INMDP_DEBUG = debug('InMemoryDP');
```
kaaax0815, line 17: `InMemoryDB` makes more sense imo

kaaax0815 (changes requested):

kaaax0815:
Very nice work.
Do you have some thoughts that will need to be done. I could create a issue tracker for that

tetuaoro:
> Do you have some thoughts that will need to be done.

Yes, what's the database provider are we going to use ? I'm going to move all the `env`s to the database first.

kaaax0815:
> Yes, what's the database provider are we going to use ? I'm going to move all the `env`s to the database first.

Good question. Have you tried out any? lowdb works on all platforms I believe

tetuaoro:
> Good question. Have you tried out any? lowdb works on all platforms I believe

Ok

src/adapters/database/inmemory.ts
```diff
@@ -0,0 +1,146 @@
...
+import type { System } from '~/ports/system/model';
+import type { User } from '~/ports/user/model';
+import { hashPassword, isPasswordStrong } from '~/server/utils/password';
+
+const DEBUG = debug('InMemoryDB');
+
+// Represent in-memory data structure
+type InMemoryData = {
+  system?: System;
+  users: Array<User>;
```
kaaax0815, line 18: Use `User[]` Syntax

src/ports/system/model.ts
```diff
@@ -0,0 +1,42 @@
+import type { SessionConfig } from 'h3';
```
kaaax0815, line 1: This File only adds typings. Maybe model and interface should be merged. As we don't follow a mvc-like pattern

src/ports/types.ts
```diff
@@ -0,0 +1,47 @@
+export type Undefined = null | undefined | 0 | '';
```
kaaax0815, line 1: Please don't do that. Undefined should be clearly defined. `null` or `undefined` depending on use case. No type coercion

src/server/api/lang.get.ts
```diff
@@ -1,4 +1,4 @@
-export default defineEventHandler((event) => {
+export default defineEventHandler(async (event) => {
```
kaaax0815, line 1: async can be removed here

src/package.json
```diff
@@ -35,6 +35,7 @@
     "qrcode": "^1.5.4",
     "tailwindcss": "^3.4.10",
     "timeago.js": "^4.0.2",
+    "uuid": "^10.0.0",
```
kaaax0815, line 38: cant find a reference to a import for this package. Should be replaced by `crypto.randomUUID`

kaaax0815 (changes requested):

kaaax0815:
I think this pr should only add the move to a database. Stuff noted in #1333 and moving all env to the ui should be done in a later stage to avoid too much work in #1333

kaaax0815:
Please move database from services and repositories into its own database folder

src/server/middleware/setup.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (
+    url.pathname.startsWith('/setup') ||
+    url.pathname === '/api/account/new'
+  ) {
+    return;
+  }
+
+  const users = await Database.getUsers();
+  if (users.length === 0) {
+    return sendRedirect(event, '/setup', 302);
```
kaaax0815, line 14: this runs on every request but is only needed once. is there a better way to do this?

src/localeDetector.ts
```diff
@@ -0,0 +1,24 @@
+// https://i18n.nuxtjs.org/docs/guide/server-side-translations
```
kaaax0815, line 1: is this really needed?

tetuaoro:
> I think this pr should only add the move to a database. Stuff noted in #1333 and moving all env to the ui should be done in a later stage to avoid too much work in #1333

Ok

kaaax0815:
i think adding a "proper" persistent database should be handled by this pr. then i'll merge it

tetuaoro:
actually, the database assume log in with a username & a password

### TODO
- [x] UI `/setup` page
- [x] timing session post
- [x] re-export types *(not required)*
- [x] compute next ip based on CIDR, not with `x` : *10.8.0.x*.

src/localeDetector.ts
kaaax0815: @tetuaoro

.gitignore
```diff
@@ -4,3 +4,5 @@
 /src/node_modules
 .DS_Store
 *.swp
+# lowdb data file
+db.json
```
kaaax0815, line 8: this shouldn't be needed as the file should be inside `/etc/wireguard`

kaaax0815:
lgtm (some thinks still can be improved but its more nitpicking than actual improvements)

tetuaoro:
> @tetuaoro What do you want to add to this repo? Because the TODO you commented are for separate prs

Yes

PR #1333, opened 2024-08-30
Rewrite in Nuxt and Typescript

kaaax0815:
## Description

This rewrites the UI and Server in Nuxt and Typescript
This moves env variables to be configurable in the ui

## Motivation and Context

Many features requested here https://github.com/wg-easy/wg-easy/discussions/1179 and in other places require better management of Database and UI.
The Database should support Atomic Writes and be typesafe.
The UI should be composed of components for reusability and easy maintenance.
The Users should be able to configure more stuff through the UI more easily.

If a stage is reached where a working baseline is reached without any regressions this will be merged. Further features will be added at a later stage

## How has this been tested?

TODO

## Screenshots (if appropriate):
TODO

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

## Tracker
- [ ] Database
- [ ] Onboarding
- [ ] Settings Page
- [ ] Move all/most env variables to ui

PR #1342, opened 2024-09-02
Refactor New UI

kaaax0815:
## Description

Refactor Code

## Types of changes

- [x] Refactor (non-breaking change which brings no user gain)

## Checklist:


kaaax0815:
@tetuaoro Request review

kaaax0815:
@tetuaoro I thought a while about migrations.
Using a default struct doesn't make that much sense imo. What if we have to change the database structure in the future? add new features?
currently it was easy by just adding a env var
But now we would have to always check if a value exists and if not return a default
maybe we could create a command or something that checks if e.g. db version is current and if not change the database to the new structure.
we could also use something like drizzle and pglite

tetuaoro:
Hi, It's all about rewriting. I'll go step by step. Optimize and UI at the end. What do mean by `default struct` ?

> What if we have to change the database structure in the future?

Return a default and let the user update in the UI, idk.

> But now we would have to always check if a value exists and if not return a default
maybe we could create a command or something that checks

Yes, we can opt that.

tetuaoro:
> Setup Middleware breaks many routes

We need a solution

tetuaoro:
Ah thank you, you've done the UI 🥰

src/server/utils/Database.ts
```diff
@@ -14,4 +13,6 @@ provider.connect().catch((err) => {
   process.exit(1);
 });
 
+// TODO: check if old config exists and tell user about migration path
```
tetuaoro, line 16: I planned to migrate the old config in the setup step

src/server/utils/config.ts
```diff
@@ -2,8 +2,6 @@ import type { SessionConfig } from 'h3';
 
 import debug from 'debug';
 
-export const PORT = process.env.PORT || '51821';
```
tetuaoro, line 5: Are you sure to remove `PORT` ?

kaaax0815:
The setup is not done. It just added some styling so the basic works.

I refactor now because temporary solutions are the most permanent ones.

Next steps. Are outlined in #1333. Most importantly good onboarding and settings ui. And moving all env vars to database

src/server/utils/Database.ts
kaaax0815: makes sense. would be a good option

src/server/utils/config.ts
kaaax0815: yes because its not used anywhere. h3 (nuxt) needs `PORT` as env var as you don't start the server yourself. `WEBUI_HOST` has to be migrated to `HOST` to work

kaaax0815:
@tetuaoro `feat-ui` is in a good state now imo. Rewriting even more (wireguard, env -> db) would be greatly appreciated

tetuaoro:
OK thank you

PR #1367, opened 2024-09-06
Fix onetimelink

AKEB:
## Description

There's a bug with the one-time link on Android TV in the Downloader app. For some reason this app does not download the link the first time. It makes two requests. That's why I made a fix that doesn't kill the one-time link immediately, but lets it live for another 10-60 seconds

## Motivation and Context

Bug fix

## How has this been tested?

On my server

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


pheiduck (approved):
Thank you @AKEB!

PR #1383, opened 2024-09-10
Feat: Improve Database Handling

kaaax0815:
## Description

This improves Database handling and DX.
Instead of praying that `#db` is set. This throws a Error message such that the frontend shows a 500 page and users know that the "Database [is] not yet initialized".
And if in the backend a function didn't wait properly for the Database (and action is not made from user) it will crash the server.
This should avoid any undefined state where e.g. migrations are not done.
The WireGuard Interface is now started after DB is connected instead of having to poll the connection state

## Motivation and Context

Fix weird issues

## How has this been tested?

Locally

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/server/middleware/session.ts
+++ b/src/server/middleware/session.ts
@@ -11,11 +11,6 @@ export default defineEventHandler(async (event) => {
     return;
   }
   const system = await Database.system.get();
-  if (!system)
-    throw createError({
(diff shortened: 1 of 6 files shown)


kaaax0815:
Just tested Build Cache, even if build fails locally the action caches to much

PR #1406, opened 2024-09-15
test: Sqlite

tetuaoro:
test sqlite arm compatible


tetuaoro:
@kaaax0815 consider using `node:sqlite` for performance improvements as `json` lacks the necessary performance if the file becomes very large. It seems `node:sqlite` works.

tetuaoro:
@kaaax0815 consider using `node:sqlite` for performance improvements as `json` lacks the necessary performance if the file becomes very large. It seems `node:sqlite` works.

kaaax0815:
`node:sqlite` is still experimental. The DB shouldn't get too big therefore I think this is a non issue.

PR #1474, opened 2024-10-15
Added random avatars support

jlnemirova:
This changes implement support of random avatars

## Description
Use environment at README


kaaax0815:
Please rework this feature against `feat-ui`

jlnemirova:
I changed a base to feat-ui.

kaaax0815:
Also the problem is that more Avatar support is planned. Like being able to select any avatar. And like sub-avatars with a icon like phone, Laptop, desktop see https://github.com/wg-easy/wg-easy/discussions/1179#discussioncomment-10264297

jlnemirova:
I see, that another user suggested a feature request for more avatar support should be incremented, but at this point I think it can be better to implement my commits, cause they don't require use of database for storing data. All avatars are generated on the fly with dicebear service api. All the data is stored is just a configuration for tunnels.

PR #1503, opened 2024-10-31
Add support advanced security settings (AmneziaWG)

kolsys:
Add support [AmneziaWG](https://github.com/amnezia-vpn/amneziawg-go) security settings 

## Description
I have add support AmneziaWG parameters to the config generator. It allows to define it via `WG_ASC_*` environment variables. See [description](https://github.com/amnezia-vpn/amneziawg-linux-kernel-module?tab=readme-ov-file#configuration).

```
WG_ASC_JC
WG_ASC_JMIN
WG_ASC_JMAX
WG_ASC_S1
WG_ASC_S2
WG_ASC_H1
WG_ASC_H2
WG_ASC_H3
WG_ASC_H4
```

Also fixed hardcoded assets path for possibility to run not only from /app path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,25 @@ These options can be configured by setting environment variables using `-e KEY="
 
 > If you change `WG_PORT`, make sure to also change the exposed port.
 
+### Advanced security settings ([AmneziaWG](https://docs.amnezia.org/documentation/amnezia-wg/) only)
+If defined `WG_ASC_JC` then advanced security settings will be enabled.
+All options are required.
+
+| Env           | Default | Example | Description                                                                                                            |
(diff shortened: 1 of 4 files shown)


kaaax0815:
This generates a config for amnezia but it still doesn't use amnezia, so I don't understand the reason for this pr

kaaax0815:
As noted in #1482 I will not merge this.
Amnezia would also have to be included in Dockerfile. Which would result in more packages in the image.

Therefore I think adding it to wg-easy is out of scope. And should be handled in a fork (which exist already).
But maybe creating a different image and clear separation of code could also be done

PR #1645, opened 2025-02-07
Feat: Account

kaaax0815:
## Tracker:
- [ ] change email, username
- [ ] change password
- [ ] change avatar
- [ ] change more?

### DIFF
--- a/src/app/components/form/PasswordField.vue
+++ b/src/app/components/form/PasswordField.vue
@@ -6,13 +6,14 @@
     :id="id"
     v-model.trim="data"
     :name="id"
-    type="text"
+    type="password"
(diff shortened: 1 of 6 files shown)


kaaax0815:
what kind of avatar support do we want? and should it be implemented with v15 or a later minor release?

kaaax0815:
the docker build takes a long time. this should be improved as it would take 15m to be able to merge a pr to check if code at least builds

Anyexyz:
Regarding the issue of adding avatars, I don't think it is very necessary at the moment. As for the translation issue, if necessary, I can provide some simplified Chinese translation help.

Issue #1559, opened 2024-12-09
[Feat]: Support for Multi-Subnet and Multi-IP Peers


amirhmoradi:
### Description

Hi, Thanks for the great work. I have a use case where a single peer (e.g., a machine or user) needs to belong to multiple subnets and be assigned multiple IP addresses. 
Implementing this now requires manual editing of the `wg0.conf` file generated by WG-Easy. I would prefer if we supported these features natively through UI or its configuration management.

I see quite a lot of progress coming for v15, where some might help in my case.

### Background:
In my WireGuard network, peers are grouped into different subnets for access control and network isolation. For example:

Subnet 1: 192.168.192.0/24 for internal tools.
Subnet 2: 192.168.194.0/24 for Kasm workspaces.
A specific machine (server-de-internal1) needs to:

* Be assigned two IP addresses: 192.168.192.2 and 192.168.194.2.
* Be part of both subnets (192.168.192.0/24 and 192.168.194.0/24).

Challenges with the Current Workflow:
*  WG-Easy does not allow assigning multiple IPs to a single peer via the web UI.
* Modifying `wg0.conf` manually works but breaks synchronization with WG-Easy.
* There is no built-in way to visualize or manage multi-subnet/multi-IP peers.

## Proposed Features
1. Multi-Subnet Support for Peers:
* Allow peers to be assigned multiple subnets (e.g., 192.168.192.0/24 and 192.168.194.0/24).
* Provide a simple UI to add subnets and manage AllowedIPs for each peer.
3. Multi-IP Assignment for Peers:
* Enable assigning multiple IP addresses to a single peer within the subnets (e.g., 192.168.192.2 and 192.168.194.2).
* Automatically reflect these changes in the generated .conf files.
4. Subnet Visualization
* Display subnets in the WG-Easy interface, showing which peers are part of each subnet.
* Optionally, allow subnet-level access control settings.
5. Persistent Configuration Management
Ensure that manual edits to wg0.conf (e.g., for advanced use cases like mine) do not get overwritten by WG-Easy or cause issues with peer synchronization.

### Alternatives

Some features might be addressed by using pfSense (but no docker support)... and some by pivpn. but i like the simplicity of wg-easy.

### Additional Context

i got help from chatgpt in writing my request, to make sure it is clear and detailed and has not many typos/errors. Happy to give more info/examples if required.

tokKurumi:
> 1 and 4 are out of scope for this project as these are extremely niche.

[...]

PR #1676, opened 2025-02-20
Feature: added multiple allowed IPs support via wg0.json

sheikhmishar:
I've added `extraAllowedIPs` and `serverPeerAllowedIPs` props to the wg0.json config file. Using those, we can manually create custom routing through multi subnets for the tunnel. Previously, it was not possible. If people are interested, I will try to add feature for configuring through web GUI

## Description
By default, we cannot assign routing for multiple subnet IPs in a tunnel via wg0.json or web GUI. I have added 2 extra properties in the json named `extraAllowedIPs` and `serverPeerAllowedIPs`. `extraAllowedIPs` adds per client extra allowed IPs which will be routed through the tunnel. `serverPeerAllowedIPs` enables server to route allowed IP traffics to the peer (i.e. client). After which, restart the docker instance by running `docker restart wg-easy` (replace wg-easy with your instance name) and afterwards keep using as it used to be.

## Motivation and Context
This change enhances the routing for multi subnet IPs without typing commands manually everytime.
[Similar Feature Request](https://github.com/wg-easy/wg-easy/issues/1559)

## How has this been tested?
1. I've first tested keeping the variables blank for backwards compatibility. It works as undefined variables are handled in the code.
2. I've tested by using my home router subnet (eg. 192.168.68.0/24) in the serverPeerAllowedIPs in one peer (i.e. client `home_gateway`), extraAllowedIPs in all other clients. That has let me access all of my devices in that subnet securely though the tunnel. No need for adding a billion peers manually.

This test has been run through directly modifying `wg.json` via `docker exec` and also from the host machine. Both works.

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ The Web UI will now be available on `http://0.0.0.0:51821`.
 
 The Prometheus metrics will now be available on `http://0.0.0.0:51821/metrics`. Grafana dashboard [21733](https://grafana.com/grafana/dashboards/21733-wireguard/)
 
-> 💡 Your configuration files will be saved in `~/.wg-easy`
+> 💡 Your configuration files will be saved in `~/.wg-easy`. You can modify `wg0.json` file for per client configuration. `extraAllowedIPs` and `serverPeerAllowedIPs` can be used for multi-network/gateway support
 
 WireGuard Easy can be launched with Docker Compose as well - just download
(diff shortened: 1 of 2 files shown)


kaaax0815:
This is already implemented with v15.

Closing this because of #1482

PR #1678, opened 2025-02-20
fixed Dockerfile HEALTHCHECK syntax

rokiden:
HEALTHCHECK options should always come before the CMD instruction

https://docs.docker.com/reference/dockerfile/#healthcheck

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN npm ci --omit=dev &&\
 # Copy build result to a new image.
 # This saves a lot of disk space.
 FROM docker.io/library/node:lts-alpine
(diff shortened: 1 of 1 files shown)


rokiden:
Now HEALTHCHECK options passed to CMD

docker inspect wgeasy
```
            "Healthcheck": {
                "Test": [
                    "CMD-SHELL",
                    "/usr/bin/timeout 5s /bin/sh -c \"/usr/bin/wg show | /bin/grep -q interface || exit 1\" --interval=1m --timeout=5s --retries=3"
                ]
            },
```

kaaax0815:
Could you please redo the pr against `feat-ui`?

Development on `master` is currently stopped (#1482) and Version 15 should be around the corner

PR #1713, opened 2025-03-06
Test: armv7 build

kaaax0815:
## Description
this is used to test this image using node lts alpine on armv7


kaaax0815:
The issue that its not building on lts is somehow fixed
but libsql still needs to support armv7

Issue #1795, opened 2025-04-05
[Bug]: Internet routing fails when multiple docker networks are present


sha0x404:
#### Description
Internet connectivity through the WireGuard VPN tunnel works inconsistently. I can always connect to the VPN and ping machines within the tunnel, but external internet access sometimes fails.
The issue appears to be related to network interface selection. Running `ip route get 8.8.8.8 | awk '{print $5}'` inside the container reveals that the interface changes between container startups (sometimes eth0, sometimes eth1, sometimes eth2).
Internet connectivity only works when the `WG_HOST` environment variable matches the interface that's actually being used for routing.

#### Environment
Using wg-easy docker image: `ghcr.io/wg-easy/wg-easy:latest`.
Docker-compose configuration with multiple networks (wireguard, adguard, proxy).

#### Steps to Reproduce

1. Deploy the container using the configuration below. (having multiple networks)
2. Connect to the VPN.
3. Restart the container several times.
4. After some restarts, internet access through the VPN stops working.

docker-compose.yaml:
```
services:
  wireguard:
    image: ghcr.io/wg-easy/wg-easy:latest
    container_name: wireguard
    restart: unless-stopped
    cap_add:
      - NET_ADMIN
      - SYS_MODULE
    sysctls:
      - net.ipv4.ip_forward=1
      - net.ipv4.conf.all.src_valid_mark=1
    environment:
      TZ: "${TZ}"
      LANG: "${LANG}"
      WEBUI_HOST: "${WEBUI_HOST}"
      PORT: "${PORT}"
      PASSWORD_HASH: "${PASSWORD_HASH}"
      UI_TRAFFIC_STATS: "${UI_TRAFFIC_STATS}"
      WG_DEVICE: "${WG_DEVICE}"
      WG_HOST: "${WG_HOST}"
      WG_PORT: "${WG_PORT}"
      WG_MTU: "${WG_MTU}"
      WG_PERSISTENT_KEEPALIVE: "${WG_PERSISTENT_KEEPALIVE}"
      WG_DEFAULT_ADDRESS: "${WG_DEFAULT_ADDRESS}"
      WG_DEFAULT_DNS: "${WG_DEFAULT_DNS}"
      WG_ALLOWED_IPS: "${WG_ALLOWED_IPS}"
    volumes:
      - "$PWD/data/wireguard/data:/etc/wireguard"
      - "/lib/modules:/lib/modules:ro"
    ports:
      - "51820:51820/udp"
    networks:
[...]

PR #1806, opened 2025-04-10
Fix: Internet routing fails when multiple docker networks are present

ddrimus:
## Description
This PR fixes the issue where internet connectivity through the WireGuard VPN tunnel works inconsistently when multiple docker networks are present. The fix ensures proper interface selection for routing traffic.

This change:
- Respects the WG_DEVICE environment variable setting
- Provides auto-detection when WG_DEVICE is set to 'auto'
- Maintains backward compatibility with existing configurations
- Ensures consistent internet access through the VPN tunnel

## Motivation and Context
The issue (#1795) occurs because the network interface selection is inconsistent when multiple docker networks are present. The interface name changes between container startups (eth0, eth1, eth2), causing internet connectivity to fail when the wrong interface is used.

## How has this been tested?
- Tested with multiple docker networks
- Verified interface selection works correctly when:
  - WG_DEVICE is not set (defaults to eth0)
  - WG_DEVICE=auto (auto-detects correct interface)
  - WG_DEVICE is set to specific interface (e.g., eth1)
- Confirmed internet connectivity remains stable across container restarts
- Tested with the exact docker-compose configuration from the issue

## Types of changes

## Checklist:


kaaax0815:
Closing because of #1482

PR #1822, opened 2025-04-18
Add weekly build workflow for Docker image deployment

maauso:
Add weekly automated Docker build workflow for security updates

## Description

Added a new GitHub Actions workflow that automatically builds and publishes a Docker image from the master branch once a week. The image is tagged with both 'weekly' and a date-based tag ('weekly-YYYY.MM.DD') for easy identification and tracking. The primary goal is to maintain images with the latest security patches applied.

## Motivation and Context

This automated weekly build ensures that Docker images are regularly updated with the latest security patches and dependencies, without requiring manual intervention. Security vulnerabilities in dependencies can be automatically addressed in these weekly builds, significantly improving the security posture of deployments between official releases.

## How has this been tested?

The workflow was tested using manual dispatch trigger to confirm correct image building and tagging. Verified the resulting images are properly tagged and accessible in the GitHub Container Registry. The weekly security-focused builds maintain all functionality while incorporating the latest security updates.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/weekly-build.yaml
+++ b/.github/workflows/weekly-build.yaml
@@ -1,0 +1,56 @@
+name: Weekly Build
+
+on:
+  schedule:
+    # Run at 00:00 on Sundays
+    - cron: '0 0 * * 0'
+  workflow_dispatch:  # Allows manual execution when needed
+
+jobs:
+  docker:
+    name: Build & Deploy Weekly Docker
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'wg-easy'
+    permissions:
+      packages: write
+      contents: read
+    steps:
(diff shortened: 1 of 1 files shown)


kaaax0815:
I don't think this is needed. Its the stability of edge but only built weekly. I don't really see the advantage

maauso:
> I don't think this is needed. Its the stability of edge but only built weekly. I don't really see the advantage

I get your point about stability, but given that a VPN is a security-critical system, wouldn't it be beneficial to at least get updated security patches?

If we're running a version of WireGuard or the base OS from two years ago, that does reduce the overall security of the solution. Or am I missing something?

PR #1829, opened 2025-04-22
Feat: distributed build

kaaax0815:
## Description

This PR distributes the docker build process across different runners. as the runners can now be the same arch as the target image and will run in parallel, the builds are way faster


kaaax0815:
from around 9m to 2m30s
and the run was without any cache

PR #1901, opened 2025-05-30
Improve documentation on password hash

lucasrattz:
Fixes broken link on deprecated warning, improve documentation for password hashing.

## Description
The console output for the deprecated PASSWORD variable was referencing a broken link. This was fixed in `src/lib/Server.js`.

Also, the content of the file wasn't properly explaining how to generate the hashed password. It was adjusted.

## Types of changes
- [x] Bug fix

## Checklist:

### DIFF
--- a/How_to_generate_an_bcrypt_hash.md
+++ b/How_to_generate_an_bcrypt_hash.md
@@ -1,22 +1,28 @@
 # Generating bcrypt-hashed password
 
-Since version 14 of wg-easy, a password hashed with bcrypt is needed instead of the plain-text password string. This doc explains how to generate the hash based on a plain-text password.
+With version 14 of wg-easy, a password hashed with bcrypt is needed instead of the plain-text password string. This doc explains how to generate the hash based on a plain-text password.
 
 ## Using Docker + node
 
-The easiest way to generate a bcrypt password hash with wgpw is using docker and node:
+- You are using docker compose
 
-```sh
-docker run ghcr.io/wg-easy/wg-easy:14 node -e 'const bcrypt = require("bcryptjs"); const hash = bcrypt.hashSync("YOUR_PASSWORD", 10); console.log(hash.replace(/\$/g, "$$$$"));'
-```
+    The easiest way to generate a bcrypt password hash with wgpw is using docker and node:
 
-The hashed password will get printed on your terminal. Copy it and use on the `PASSWORD_HASH` environment variable in your docker compose.
+    ```sh
+    docker run ghcr.io/wg-easy/wg-easy:14 node -e 'const bcrypt = require("bcryptjs"); const hash = bcrypt.hashSync("YOUR_PASSWORD", 10); console.log(hash.replace(/\$/g, "$$$$"));'
+    ```
 
-If you are using `docker run` instead of docker compose for running wg-easy, you must enclose the hash string in single quotes (`'...'`) and unescape the dollar signs. You can use this command instead:
+    The hashed password will get printed on your terminal. Copy it and use on the `PASSWORD_HASH` environment variable in your docker compose.
 
-```sh
-docker run --rm ghcr.io/wg-easy/wg-easy:14 node -e "const bcrypt = require('bcryptjs'); const hash = bcrypt.hashSync('YOUR_PASSWORD', 10); console.log('\'' + hash + '\'');"
-```
+- You are using `docker run`
+
+    If you are using `docker run` for running wg-easy, you must enclose the hash string in single quotes (`'...'`). You can use this command:
+
+    ```sh
+    docker run --rm ghcr.io/wg-easy/wg-easy:14 node -e "const bcrypt = require('bcryptjs'); const hash = bcrypt.hashSync('YOUR_PASSWORD', 10); console.log('\'' + hash + '\'');"
+    ```
+
+    The hashed password will get printed on your terminal. Copy it and use on the `PASSWORD_HASH` environment variable in your docker run command.
 
 ## Using Docker + wgpw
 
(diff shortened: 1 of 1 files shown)


src/lib/Server.js
```diff
@@ -308,7 +308,7 @@ module.exports = class Server {
     );
 
     if (PASSWORD) {
-      throw new Error('DO NOT USE PASSWORD ENVIRONMENT VARIABLE. USE PASSWORD_HASH INSTEAD.\nSee https://github.com/wg-easy/wg-easy/blob/master/How_to_generate_an_bcrypt_hash.md');
+      throw new Error('DO NOT USE PASSWORD ENVIRONMENT VARIABLE. USE PASSWORD_HASH INSTEAD.\nSee https://github.com/wg-easy/wg-easy/blob/production/How_to_generate_an_bcrypt_hash.md');
```
kaaax0815, line 311: should now be v14

How_to_generate_an_bcrypt_hash.md
```diff
@@ -1,28 +1,33 @@
...
-- Generate bcrypt password hashes.
-- Easily integrate with `wg-easy` to enforce password requirements.
+The easiest way to generate a bcrypt password hash with wgpw is using docker and node:
 
-## Usage with Docker
+```sh
+docker run ghcr.io/wg-easy/wg-easy node -e 'const bcrypt = require("bcryptjs"); const hash = bcrypt.hashSync("YOUR_PASSWORD", 10); console.log(hash.replace(/\$/g, "$$$$"));'
+```
+
+The hashed password will get printed on your terminal. Copy it and use on the `PASSWORD_HASH` environment variable.
```
kaaax0815, line 13: does escaping the dollar signs properly work when using docker run?
does it work when using an env file?

kaaax0815 (changes requested):

src/lib/Server.js
lucasrattz: Addressed

How_to_generate_an_bcrypt_hash.md
lucasrattz: I've added information on the docker run. I'm not sure about the env file. Maybe someone who uses it can add this information in a later PR, given the current text also doesn't have it.

How_to_generate_an_bcrypt_hash.md
```diff
@@ -1,28 +1,35 @@
-# wg-password
+# Generating bcrypt-hashed password
 
-`wg-password` (wgpw) is a script that generates bcrypt password hashes for use with `wg-easy`, enhancing security by requiring passwords.
+Since version 14 of wg-easy, a password hashed with bcrypt is needed instead of the plain-text password string. This doc explains how to generate the hash based on a plain-text password.
```
kaaax0815, line 3: i would clarify that only v14 has this problem. currently it would suggest that v15 suffers the same problem

How_to_generate_an_bcrypt_hash.md
```diff
@@ -1,28 +1,35 @@
...
+The easiest way to generate a bcrypt password hash with wgpw is using docker and node:
 
-## Usage with Docker
+```sh
+docker run ghcr.io/wg-easy/wg-easy node -e 'const bcrypt = require("bcryptjs"); const hash = bcrypt.hashSync("YOUR_PASSWORD", 10); console.log(hash.replace(/\$/g, "$$$$"));'
+```
+
+The hashed password will get printed on your terminal. Copy it and use on the `PASSWORD_HASH` environment variable.
+
+If you are using `docker run` for running wg-easy, you must enclose the hash string in single quotes (`'...'`).
```
kaaax0815, line 15: this is only partially correct as with docker run you can't double the dollar sign

How_to_generate_an_bcrypt_hash.md
```diff
@@ -1,28 +1,35 @@
...
+
+## Using Docker + wgpw
 
-To generate a bcrypt password hash using docker, run the following command :
+`wg-password` (wgpw) is a script that generates bcrypt password hashes. You can use it with docker:
 
 ```sh
 docker run ghcr.io/wg-easy/wg-easy wgpw YOUR_PASSWORD
```
kaaax0815, line 22: tag version here too

How_to_generate_an_bcrypt_hash.md
```diff
@@ -1,28 +1,35 @@
...
-b2
-$ echo "$2b$12$coPqCsPtcF"
-b2
-$ echo '$2b$12$coPqCsPtcF'
-$2b$12$coPqCsPtcF
+```sh
+PASSWORD_HASH='$2b$12$coPqCsPtcFO.Ab99xylBNOW4.Iu7OOA2/ZIboHN6/oyxca3MWo7fW'
+```
+
+In this example, the `$2b$12$coPqCsPtcFO.Ab99xylBNOW4.Iu7OOA2/ZIboHN6/oyxca3MWo7fW` string is your hashed password. For using it with docker-compose, you need to escape the `$` characters by adding another `$` before them, or they will get interpreted as variables. The final password you can use in docker-compose will look like this:
```
kaaax0815, line 31: clarify that you need to each a dollar sign before **each** one. as this was a problem in the past where users would only double the first one

kaaax0815 (changes requested):

How_to_generate_an_bcrypt_hash.md
lucasrattz: IMO this is already clear. The dollar signs are talked about in plural, and there is an example below showing that.

How_to_generate_an_bcrypt_hash.md
lucasrattz: This documentation will only be shown to users on the v14, as it is not in the main branch, nor indexed in the new docs. Hence, I don't think this distinction is useful.

How_to_generate_an_bcrypt_hash.md
lucasrattz: Addressed

How_to_generate_an_bcrypt_hash.md
lucasrattz: Done

kaaax0815:
i think it would more sense to use the https://github.com/wg-easy/wg-easy/blob/c6dce0f6fb2e28e7e40ddac1498bd67e9bb17cba/How_to_generate_an_bcrypt_hash.md guide from the master again which was removed due to v15

kaaax0815:
@lucasrattz any objections?

lucasrattz:
> Quote reply

No, I think the objective is to link proper documentation in the depraction warning. As long as that's done I'm ok with it.

kaaax0815:
I meant the commit. If you are ok with merging

kaaax0815 (approved):

PR #1948, opened 2025-06-10
Add support for armv6

cgm616:
By working around some problems in the Dockerfile, this change adds armv6 support back. **NOTE:** these changes are brittle and probably not acceptable. The goal of this PR is to detail how I got it to work in case a better version can be upstreamed.

## Description
- Due to an [outstanding bug](https://github.com/nodejs/docker-node/issues/1946) in Node's docker container on armv6, this version of the Dockerfile uses `node:18-alpine3.21` instead of `node:lts-alpine`. _This is not desirable and I'm looking for another workaround._
- `node-argon2` does not ship prebuilt binaries for armv6. This causes runtime errors unless we rebuild `argon2` for the correct platform. _This should be restricted to rebuilding binaries when they don't already exist—I'm not super familiar with `node-gyp` and I don't know how to do that._

## Motivation and Context
v15 of wg-easy removed armv6 support, but not because of any deep problems. That made it impossible to use wg-easy on Raspberry Pi 1 devices (like the A, B, B+, etc.).

## How has this been tested?
I tested these changes by running a fully-working wg-easy instance built this way on a Raspberry Pi B+.

## Types of changes

## Checklist:


kaaax0815:
Node 18 reached EOL. Therefore this is not an option. And Node 22 does not support armv6

cgm616:
I've moved the Dockerfile back to `node:lts-alpine` and while it does not work locally (Intel-based Mac), the docker build is successful on GitHub Actions in all three original architectures + armv6. See [commit](https://github.com/cgm616/wg-easy/commit/fb6d51e0df5b9e93817871e9c3699ffad050cfdc) and [run](https://github.com/cgm616/wg-easy/actions/runs/15577275819/job/43864484796).

I've pushed the changes up to `cgm616/wg-easy` on branch `armv6` (the branch backing this pull request). If you re-open the PR and approve the checks, they should work.

The only thing left to fix is the `argon2` rebuild, which only needs to happen on `armv6`.

kaaax0815:
You can't even download nodejs for armv6 for a long time.
this just adds time to ci and more work for the maintainers.

PR #1951, opened 2025-06-11
Add option to disable ipv6

kaaax0815:
## Description

This adds an option to disable IPv6.
This will only be documented in the docs, as this should only be used as a last resort.

## Motivation and Context

Many Synology users can't run v15

## Additional Information

This is currently released under the `:development` tag behind the `DISABLE_IPV6` env var


kaaax0815:
This is currently released under the `:development` tag

PR #1977, opened 2025-06-22
Add custom site title feature with database migration

Mavrag:
This PR adds the ability to customize the site title from the Admin Panel.

Changes:
- Added a `siteTitle` field to the General table schema
- Created a migration file to add the column to existing databases
- Added a text input field in the Admin Panel settings
- Updated the app.vue to use the custom title with a fallback to "WireGuard"
- Fixed data refresh logic for reactive updates

This enhancement allows administrators to customize the browser tab title for their WireGuard Easy installation.

### DIFF
--- a/src/app/app.vue
+++ b/src/app/app.vue
@@ -19,6 +19,12 @@ toast.setToast(toastRef);
 // make sure to fetch release early
 useGlobalStore();
 
+// Fetch site title from general configuration with a key for refreshing
+const { data: generalConfig } = await useFetch('/api/admin/general', { key: 'site-title' });
+
+// Compute the title with fallback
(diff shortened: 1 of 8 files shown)


Mavrag:
I added a small but important improvement to ensure the browser tab title updates immediately after changing the site title in the admin panel. This uses Nuxt's refreshNuxtData API to refresh the title data without requiring a full page reload.

kaaax0815:
Your text in this pr and the feature request seem like AI. please avoid that.

also please fix the formatting issues and generate the migrations using drizzle.  see `db:generate` in `src/package.json`. 
and maybe it makes sense to make site title not null and specify a default

PR #1997, opened 2025-07-01
Improve Chinese translation and fix 4-character Chinese suggestion button

yhl452493373:
+ Improve Chinese translation of suggest
+ Fix 4-character Chinese suggestion button

Before
屏幕截图_1-7-2025_201648_103 40 13 68

After
屏幕截图_1-7-2025_20240_103 40 13 68


src/app/components/Form/HostField.vue
```diff
@@ -21,14 +21,20 @@
...
     </ClientOnly>
   </div>
 </template>
 
+<style scoped lang="css">
+.w-16{
+  width: 4rem;
+}
+</style>
+
```
kaaax0815: the style tag should be removed as tailwind is used for styling

kaaax0815 (changes requested):

src/app/components/Form/HostField.vue
kaaax0815: did you try out other languages with this css fix? does it break other languages?

yhl452493373: I modified the code and tested each language this time.

kaaax0815: Thank you. But hard coding every value is not a good practice. I'll have to look into it

yhl452493373: Maybe you can find a better way.

kaaax0815: `whitespace-nowrap` should do the job

yhl452493373: I test this, it's works.

kaaax0815:
You changed the translation. I can't read chinese. are these changes needed anyway or should they be reverted?

yhl452493373:
> You changed the translation. I can't read chinese. are these changes needed anyway or should they be reverted?

This make the panel items name mentioned in the description matches the name on the left. Just merge directly, no need to revert.

kaaax0815:
There is a formatting issue that needs to be fixed

PR #2048, opened 2025-07-22
Add interface_name to the wg network in docker-compose.yml

mndrei:
Added `interface_name: eth0` to the `wg` network in the example docker-compose.yml

## Description
When using multiple networks (like with Traefik or caddy-docker-proxy), the interface name keeps changing after a container restart, sometimes preventing the VPN from working. Adding an explicit name for the main `wg` network prevents them from switching with other networks (ie: going from `eth0` to `eth1` and vice versa)

## Motivation and Context

I was having issues with my current wg-easy setup, sometimes preventing any network traffic through the VPN but still successfully connecting.

## How has this been tested?

 I tested my setup without using caddy-docker-proxy, and had no issues. With multiple networks, running `docker exec wg-easy ip a` shows both networks under `eth0` and `eth1`, with them swapping sometimes after a container restart. With the changes to my compose.yaml, running the above command now shows the `wg` network always on `eth0`

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation. (Maybe, works by default but might be good to mention under the reverse proxy setups)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,7 @@ services:
     container_name: wg-easy
     networks:
       wg:
+        interface_name: eth0
         ipv4_address: 10.42.42.42
         ipv6_address: fdcc:ad94:bacf:61a3::2a
     volumes:


kaaax0815:
based on https://github.com/wg-easy/wg-easy/issues/1795#issuecomment-2880052214 this only works since docker compose v2.36.0. i am running on almalinux 9.6, which is fairly new, and is still on v2.35.1

and like you mentioned i think it will error if you don't add interface_name to every interface

the old docker compose version seems to be a docker, rhel and arm64 issue

mndrei:
Ah, I should of checked for other versions of docker compose. I didn't realize since I'm on a rolling release distro. 

It does still work with only the `wg` network having an interface name in my current compose though.

I'll close this for now. maybe a change to the docs might help, but it's an issue that only really affect those who are probably writing their own compose anyways.

Thanks for your work on the project, I've been using it constantly since v14!

PR #2052, opened 2025-07-24
Feat: Allow empty dns

### LINKED ISSUE #1907: [Bug]: v15: I cannot omit WireGuard DNS parameter
### Describe the bug

Prior to wg-easy v15, I was able to set the WG_DEFAULT_DNS environment variable to an empty value. This caused the DNS parameter to be omitted from the generated client configuration, which was the desired behavior in my case.

I want WireGuard to respect my client machine’s existing DNS setup and not override it with a DNS setting from the server.

However, in wg-easy v15, using the Admin Panel:
 • When I remove the last DNS entry, I get an error and the change is not saved.
 • This means I’m now unable to leave the DNS field empty, which effectively forces a DNS configuration on my clients.

Is there currently any other way to leave the DNS parameter empty in the configuration? Or could you consider reintroducing the ability to omit the DNS parameter entirely?

### Expected behavior

I can set DNS parameter empty.

kaaax0815:
## Description

This PR allows to have no DNS in the User Config

But: You cannot set no DNS for only one Device. As this would conflict with the fallback behavior to the global config

Note: this will result in an empty dns field. `DNS = `. But this seems to work in the official wireguard clients

## Motivation and Context

Closes #1907 

## How has this been tested?

Locally

## Types of changes

## Checklist:

### DIFF
--- a/src/server/utils/types.ts
+++ b/src/server/utils/types.ts
@@ -80,7 +80,6 @@ export function validateZod<T>(
             if (t) {
               let newMessage = null;
               if (v.message.startsWith('zod.')) {
-                console.log(v);
                 switch (v.code) {
                   case 'too_small':
                     switch (v.origin) {


GideonBear (approved):

GideonBear:
```diff
4c4
< DNS = 1.1.1.1, 2606:4700:4700::1111
---
> DNS = 
```

Works!

> Note: this will result in an empty dns field. DNS = . But this seems to work in the official wireguard clients

Yes, I have been using this both in wg-easy v14, and wg-easy v15 by manually changing the .conf file.

sckiyo:
The official windows wireguard client refuses to accept a configuration file with an empty DNS field. "DNS = " should preferably be removed if set no DNS.

GideonBear:
> The official windows wireguard client refuses to accept a configuration file with an empty DNS field. "DNS = " should preferably be removed if set no DNS.

I haven't checked the config file (I deleted it and now no longer use wg on Windows), but I know that with v14 it worked fine on the official Windows client; although I don't know if that had an empty or missing DNS field.

sckiyo:
> > The official windows wireguard client refuses to accept a configuration file with an empty DNS field. "DNS = " should preferably be removed if set no DNS.
> 
> I haven't checked the config file (I deleted it and now no longer use wg on Windows), but I know that with v14 it worked fine on the official Windows client; although I don't know if that had an empty or missing DNS field.

Yes, in v14, if there is no DNS, there will be no DNS field. However, if there is "DNS = " in the configuration file but no DNS after it, the Windows client will pop up a window to prompt and reject it, but in Linux it seems to work fine.

kaaax0815:
fixed in #2057

PR #2077, opened 2025-08-05
feat: Add Português Brasil translation

ciro-mota:
## Description
Add i18n Português Brasil translation.

## Motivation and Context
Simply want to contribute to another language translation.

## How has this been tested?
This is not tested in my local yet.

## Types of changes

## Checklist:

### DIFF
--- a/src/i18n/i18n.config.ts
+++ b/src/i18n/i18n.config.ts
@@ -7,6 +7,7 @@ import zhhk from './locales/zh-HK.json';
 import zhcn from './locales/zh-CN.json';
 import ko from './locales/ko.json';
 import es from './locales/es.json';
+import ptbr from './locales/pt-BR.json';
 
 export default defineI18nConfig(() => ({
   legacy: false,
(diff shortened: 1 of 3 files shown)


kaaax0815:
Are there significant differences between the various Portuguese dialects spoken in different countries? If not, I would prefer one general translation

ciro-mota:
Sorry for the delay.

Yes, there are differences between the Portuguese spoken in Portugal and other countries where Portuguese is spoken in Brazil. In Brazil, we accept more English terms in our vocabulary without translation, such as "Download" and "Upload." "Endpoint" is another word that anyone using the application should know.

There are other differences, for example, "password" in Portuguese from Portugal is called "palavra passe," while in Brazilian Portuguese we simply call it "senha."

In any case, I can try to create a unified translation for both.

kaaax0815:
If it doesn't make sense for a unified translation. there is no problem with two.
i was just asking to make the translations easier for future updates

ciro-mota:
Ideally, there should be two, just like with other software/projects.

kaaax0815:
Thank you very much

Issue #2053, opened 2025-07-24
[Feat]: AmneziaWG


cany748:
### Description

I know this question has been opened before, but I suggest taking a look at it again.
An example implementation based on the current wg-easy codebase is here: [https://github.com/cany748/awg-easy](https://github.com/cany748/awg-easy). Most of the changes will not be required for the original repository (changes in links and names).

### Suggested Solution

I see several implementations of this functionality:
1) Use 1 container for both options. When starting the container, check if the user has `awg` installed, if so, use it and show additional fields in the application. The downside is the larger container size, since the image does not use Alpine.
2) Use 2 different containers. One with `awg`, the other with the classic `wg`. The downside is the need to maintain two separate images.

kaaax0815:
is there no way to install amnezia on alpine?

but i think a good maintained fork would be the best option. if it doesn't diverge to much

cany748:
> is there no way to install amnezia on alpine?

There is a possibility, but then you will have to build `amneziawg-tools` manually. It is not very difficult, you can use the container used to build the application for this.

> but i think a good maintained fork would be the best option. if it doesn't diverge to much

Using a fork will divide users and confuse them when problems arise. People in Russia, China, and Iran need AWG. 

I can start small and add experimental support with small changes.

thiswillbeyourgithub:
Very interested in this, keep us posted please :)

kaaax0815:
@cany748 if you find a way to incorporate both wg and awg. i will gladly look into it

kaaax0815:
@cany748 Any update?

cany748:
> Any update?

I think it's worth waiting for full support of AWG 2.0

it should be soon

kaaax0815:
@cany748 is there any reference to awg 2.0 other than some github threads? some official spec or anything?

cany748:
> [@cany748](https://github.com/cany748) is there any reference to awg 2.0 other than some github threads? some official spec or anything?

https://docs.amnezia.org/documentation/amnezia-wg/


Windows, Apple and Android clients already support AWG 2.0, but new versions are not yet available in stores.
Go implementation is ready, but Linux kernel module [is not ready yet](https://github.com/amnezia-vpn/amneziawg-linux-kernel-module/pull/88)

documentation should be ready after the implementation at the kernel level is completed

freeman1111:
Any update?

kaaax0815:
[...]

PR #2102, opened 2025-08-12
feat: add amneziawg support

cany748:
## Description
This PR implements: installing amneziawg-tools and defining the wireguard-tools executable. 
If the user has the amneziawg kernel module, it will be used instead of the regular wg.
This PR does not add anything to the UI or DB.
partially solves #2053

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,12 @@ RUN pnpm install
 COPY src ./
 RUN pnpm build
 
+# Build amneziawg-tools
+RUN apk add linux-headers build-base git && \
+    git clone https://github.com/amnezia-vpn/amneziawg-tools.git && \
+    cd amneziawg-tools/src && \
+    make
+
 # Copy build result to a new image.
 # This saves a lot of disk space.
 FROM docker.io/library/node:lts-alpine
@@ -32,6 +38,10 @@ RUN cd /app/server && \
 # cli
 COPY --from=build /app/cli/cli.sh /usr/local/bin/cli
 RUN chmod +x /usr/local/bin/cli
+# Copy amneziawg-tools
+COPY --from=build /app/amneziawg-tools/src/wg /usr/bin/awg
+COPY --from=build /app/amneziawg-tools/src/wg-quick/linux.bash /usr/bin/awg-quick
+RUN chmod +x /usr/bin/awg /usr/bin/awg-quick
 
 # Install Linux packages
 RUN apk add --no-cache \
@@ -43,6 +53,9 @@ RUN apk add --no-cache \
     kmod \
     iptables-legacy \
     wireguard-tools
+
+RUN mkdir -p /etc/amnezia
+RUN ln -s /etc/wireguard /etc/amnezia/amneziawg
 
 # Use iptables-legacy
 RUN update-alternatives --install /usr/sbin/iptables iptables /usr/sbin/iptables-legacy 10 --slave /usr/sbin/iptables-restore iptables-restore /usr/sbin/iptables-legacy-restore --slave /usr/sbin/iptables-save iptables-save /usr/sbin/iptables-legacy-save
--- a/src/nuxt.config.ts
+++ b/src/nuxt.config.ts
@@ -95,7 +95,7 @@ export default defineNuxtConfig({
     esbuild: {
       options: {
         // to support big int
-        target: 'es2020',
+        target: 'node20',
       },
     },
     alias: {
--- a/src/server/utils/config.ts
+++ b/src/server/utils/config.ts
@@ -12,6 +12,8 @@ export const OLD_ENV = {
   PASSWORD_HASH: process.env.PASSWORD_HASH,
 };
 
+const OVERRIDE_AUTO_AWG = process.env.OVERRIDE_AUTO_AWG?.toLowerCase();
+
 export const WG_ENV = {
   /** UI is hosted on HTTP instead of HTTPS */
   INSECURE: process.env.INSECURE === 'true',
@@ -21,11 +23,10 @@ export const WG_ENV = {
   DISABLE_IPV6: process.env.DISABLE_IPV6 === 'true',
   /** Override automatic detection */
   OVERRIDE_AUTO_AWG:
-    process.env.OVERRIDE_AUTO_AWG === 'true'
-      ? true
-      : process.env.OVERRIDE_AUTO_AWG === 'false'
-        ? false
-        : undefined,
+    OVERRIDE_AUTO_AWG === ('wg' as const) ||
+    OVERRIDE_AUTO_AWG === ('awg' as const)
+      ? OVERRIDE_AUTO_AWG
+      : undefined,
   /** TODO: delete on next major version */
   EXPERIMENTAL_AWG: process.env.EXPERIMENTAL_AWG === 'true',
 };
--- a/src/server/utils/config.ts
+++ b/src/server/utils/config.ts
@@ -19,6 +19,15 @@ export const WG_ENV = {
   PORT: assertEnv('PORT'),
   /** If IPv6 should be disabled */
   DISABLE_IPV6: process.env.DISABLE_IPV6 === 'true',
+  /** Override automatic detection */
+  AWG:
+    process.env.AWG === 'true'
+      ? true
+      : process.env.AWG === 'false'
+        ? false
+        : undefined,
+  /** TODO: delete on next major version */
+  EXPERIMENTAL_AWG: process.env.EXPERIMENTAL_AWG === 'true',
 };
 
 export const WG_INITIAL_ENV = {
--- a/src/server/utils/config.ts
+++ b/src/server/utils/config.ts
@@ -20,10 +20,10 @@ export const WG_ENV = {
   /** If IPv6 should be disabled */
   DISABLE_IPV6: process.env.DISABLE_IPV6 === 'true',
   /** Override automatic detection */
-  AWG:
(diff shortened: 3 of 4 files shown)


src/nuxt.config.ts
```diff
@@ -95,7 +95,7 @@ export default defineNuxtConfig({
     esbuild: {
       options: {
         // to support big int
-        target: 'es2020',
+        target: 'node20',
```
kaaax0815, line 98: any reason for this change?

Dockerfile
```diff
@@ -32,6 +38,10 @@ RUN cd /app/server && \
 # cli
 COPY --from=build /app/cli/cli.sh /usr/local/bin/cli
 RUN chmod +x /usr/local/bin/cli
+# Copy amneziawg-tools
+COPY --from=build /app/amneziawg-tools/src/wg /usr/bin/awg
+COPY --from=build /app/amneziawg-tools/src/wg-quick/linux.bash /usr/bin/awg-quick
+RUN chmod +x /usr/bin/awg /usr/bin/awg-quick
```
kaaax0815, line 44: how much does this increase the container size?

src/server/utils/wgHelper.ts
```diff
@@ -9,6 +9,10 @@ type Options = {
   enableIpv6?: boolean;
 };
 
+const wgExecutable = await exec('modinfo amneziawg')
+  .then(() => 'awg' as const)
+  .catch(() => 'wg' as const);
```
kaaax0815: this should be set manually or have an override of some sort

src/nuxt.config.ts
cany748: Top level await.
The container uses node.js version 22. All versions less than 20 EOL.

Dockerfile
cany748: ≈ 150kb

kaaax0815: | Description     | Size (MB) | Difference (MB) | Percentage Increase |
|-----------------|-----------|-----------------|---------------------|
| without    | 353.15    | -               | -                   |
| with awg   | 353.44    | 0.29            | 0.08%               |

Therefore i think the feature makes sense to add directly

src/server/utils/wgHelper.ts
cany748: This can be done using an environment variable. Suggest me a name for the variable and I will implement it

kaaax0815: depends on the plan forward.
should this be configured through env? or the ui?
what happens if one chooses awg but does not have the kernel module installed? fallback to wg? error?
what if one has awg but not wg loaded, and did not change anything? fallback to awg?

iirc if no options are set for awg (which is currently the case) its compatible with wg. correct?

the plan forward for the ui should be pretty simple. if awg is selected show the new options in the client edit, if not then not. and also output the options if awg is selected or not.

what about setting defaults for the awg options in the admin panel? should these be displayed all the time? or be hidden if not awg?

kaaax0815: > should this be configured through env? or the ui?

i prefer the ui. but as awg is inherently a advanced setup, env would make sense too

> what happens if one chooses awg but does not have the kernel module installed? fallback to wg? error?
what if one has awg but not wg loaded, and did not change anything? fallback to awg?

no fallback, always error would be the obvious choice (everything else could result in confusion why e.g. the ui fields are not visible). or it should be clear that the flag is only the preferred option

> what about setting defaults for the awg options in the admin panel? should these be displayed all the time? or be hidden if not awg?

only be hidden. if one switched from awg>wg>awg the old changes are preserved

cany748: Automatic detection should work correctly in 99% of cases. So the setting will be an experimental override. The user must clearly understand why he needs it.
AWG doesn't make sense without additional parameters, which we will add later.
If you try to run the config with additional parameters in a regular WG, you will get an error. But any correct config of a regular WG will work in AWG.

So I think we should only handle 2 states: awg is present or not, no edge cases. This will simplify the code base and make it easier to maintain.

p.s. sorry for my english, i use translator

cany748: > > should this be configured through env? or the ui?
> 
> i prefer the ui. but as awg is inherently a advanced setup, env would make sense too
> 
> > what happens if one chooses awg but does not have the kernel module installed? fallback to wg? error?
> > what if one has awg but not wg loaded, and did not change anything? fallback to awg?
> 
> no fallback, always error would be the obvious choice (everything else could result in confusion why e.g. the ui fields are not visible). or it should be clear that the flag is only the preferred option
> 
> > what about setting defaults for the awg options in the admin panel? should these be displayed all the time? or be hidden if not awg?
> 
> only be hidden. if one switched from awg>wg>awg the old changes are preserved

I completely agree with all points

kaaax0815: > I completely agree with all points

You can't agree with questions😂

---
To summarize.

On container boot we check if awg is available. if yes awg used. if false wg is used. if the override is set to wg, wg is used even if awg is available. (a override to awg makes no sense, as its detected automatically anyway). as this is advanced, the override is set through a env var (and most users don't have awg installed anyway) (but the sideeffect that if awg installed it will be used should be documented)

a field in the admin panel should display if awg or wg is currently used, and this should be displayed in the logs.

if awg is used, the client edit page shows more fields
if awg is used, the admin panel has a new site for setting awg options (or a new section in the Config site)

this feature could be considered a breaking change, as it could use awg on systems where it previously didn't.

this could be handled in two ways. the default options for awg are such that they are compatible with wg. or you we throw away the automatic detection

cany748: > a override to awg makes no sense, as its detected automatically anyway

This makes sense if someone wants to use the userspace implementation, but it will require rebuilding the container anyway.

> this could be handled in two ways. the default options for awg are such that they are compatible with wg. or you we throw away the automatic detection

I propose to move automatic detection to a separate variable (for example EXPERIMENTAL_AWG), when setting this variable we will enable automatic detection. By default this variable will be false, and in the next major version (16) we can remove it and make the value true.

kaaax0815: > I propose to move automatic detection to a separate variable (for example EXPERIMENTAL_AWG), when setting this variable we will enable automatic detection. By default this variable will be false, and in the next major version (16) we can remove it and make the value true.

lgtm. would you implement that?

cany748: > lgtm. would you implement that?

Done

src/server/utils/config.ts
```diff
@@ -19,6 +19,15 @@ export const WG_ENV = {
   PORT: assertEnv('PORT'),
   /** If IPv6 should be disabled */
   DISABLE_IPV6: process.env.DISABLE_IPV6 === 'true',
+  /** Override automatic detection */
+  AWG:
+    process.env.AWG === 'true'
+      ? true
+      : process.env.AWG === 'false'
```
kaaax0815: currently if a variable is not set it defaults to a default value.
here not setting it has its own meaning. therefore i think it should be named something like `OVERRIDE_AUTO_AWG` or something. can't think of a better name rn

cany748: Done

src/server/utils/config.ts
```diff
@@ -19,6 +19,15 @@ export const WG_ENV = {
...
   DISABLE_IPV6: process.env.DISABLE_IPV6 === 'true',
+  /** Override automatic detection */
+  OVERRIDE_AUTO_AWG:
+    process.env.OVERRIDE_AUTO_AWG === 'true'
+      ? true
+      : process.env.OVERRIDE_AUTO_AWG === 'false'
+        ? false
+        : undefined,
+  /** TODO: delete on next major version */
+  EXPERIMENTAL_AWG: process.env.EXPERIMENTAL_AWG === 'true',
```
kaaax0815: my bad. the renaming now makes the logic a bit different, as setting override to false would imply that it just does nothing.
- rename back
- or the values would be 'awg' 'wg' or undefined

cany748: Done

kaaax0815:
@cany748 see #2108 if you have any corrections

z0xca:
Thx! This will hopefully make it so people migrate back to this repo from all the amnezia-wg-easy forks. Although I can't seem to grasp how to use it, will deploying with the `edge` tag automatically use amneziawg-go instead of wireguard-go?

kaaax0815:
There is neither WireGuard-go nor amnezia-go support. You have to install the kernel module.

LucAce:
A recommendation and a question,

1) The amneziawg-tools repository has been adding features and then reverting some of those features in the last couple months.  I recommend pinning the clone of that repository that is done in the dockerfile to a specific revision to make the image builds more reproducible and auditable.
https://github.com/wg-easy/wg-easy/blob/ef463d3d8540554e2d0a147e84f64b116382480a/Dockerfile#L19C15-L19C61

2) Since amneziawg-tools is replacing some of the wireguard userpace applications, have the changes that amneziawg-tools will introduce been evaluated from a security perspective?

cany748:
> 1. I recommend pinning the clone of that repository that is done in the dockerfile to a specific revision.

yes we will do it



> 2\. Since amneziawg-tools is replacing some of the wireguard userpace applications, have the changes that amneziawg-tools will introduce been evaluated from a security perspective?

amneziawg-tools will not replace anything, but will be installed next to the original wireguard-tools. If the user does not have the kernel module installed, AWG will never run.

Issue #2152, opened 2025-09-07
[Feat]: Available on DockerHub?


litetex:
### Description

Is the Docker/Containerimage also available somewhere on DockerHub (a quick search yielded no results) or somewhere else besides GitHub?

Asking because [ghcr.io only supports IPv4](https://github.com/orgs/community/discussions/10539) and I would like to deploy the image on an IPv6 only host...

### Suggested Solution

Deploy the image additionally to DockerHub

### Alternatives

Deploy the image to another IPv6 registry, like https://codeberg.org/

kaaax0815:
I do hate github for not adding ipv6 support.

I do like using codeberg, but i am not familiar with its docker support. could i push to the codeberg registry using github actions? looks like it. but not the most important thing rn. would appreciate a pr

litetex:
I will create a PR with a Proof of Concept that adds support for DockerHub, however you can also replace it with something else if you want or additionally add another registry (e.g Codeberg).

Adding support for another container registry is quite easy:
1. Login to said registry, e.g. with [docker/login-action](https://github.com/docker/login-action)
2. Specify the images (including registry) in [docker/metadata-action](https://github.com/docker/metadata-action)

However you need to create accounts and projects/repos first on DockerHub otherwise this will not work at all.

PS: In the meantime I manually mirrored the image myself: https://hub.docker.com/r/litetex/ghcr.wg-easy.wg-easy

litetex:
Btw is there any specific reason why you build on different machines with differnt architectures and then merge the results?
I think you can do the same on a single machine with something like buildkit.

litetex:
PoC is available at https://github.com/wg-easy/wg-easy/pull/2158

kaaax0815:
> Btw is there any specific reason why you build on different machines with differnt architectures and then merge the results?
> I think you can do the same on a single machine with something like buildkit.


Yes. It greatly reduces the build time from around 8m to 2m

kaaax0815:
see https://codeberg.org/wg-easy/-/packages/container/wg-easy/edge

kaaax0815:
it looks like we are already using 816 MiB / 1,5 GiB of the "griesel" storage (what ever that means). i'll have to look into the limits

litetex:
> it looks like we are already using 816 MiB / 1,5 GiB of the "griesel" storage (what ever that means). i'll have to look into the limits

[...]

PR #2158, opened 2025-09-10
PoC showcasing different registries

litetex:
PoC for #2152


kaaax0815:
i would prefer keeping the parallel build, remove extra compression, and the mode min cache is better invalidation, and remove extra provenance as the docker action does that by default

litetex:
Well this is just a PoC that showcases how this _could_ be done. 
Feel free to adopt the changes to your needs :)

kaaax0815:
not needed, as it doesn't help with the parallel workflow

PR #1978, opened 2025-06-22
Add client search feature

Mavrag:
🚀 This PR adds a **client search feature** to the WireGuard Easy interface.

firefox_QvCGQkIJey

Closes #1848

### ✨ Features Implemented:

- 🔍 Added search functionality to filter clients by name
- 🧩 Created a new `ClientsSearch.vue` component with search input and clear button
- 📦 Updated the Pinia store with search state and filtered clients logic
- 🖥️ Modified the client list to display filtered results
- 🌐 Added i18n support for the search placeholder text
- 🎨 Ensured proper layout and styling in the header section

The search is **reactive** and updates the client list in real-time as the user types, providing a smooth and intuitive user experience. 🚀

Let me know if you’d like me to tweak anything!

PR #2170, opened 2025-09-18
Add search / filter box

yuWorm:
## Description

Add a filter box to the client list and provide i18n translation for for its placeholder. base  (#1978)

## Motivation and Context

When there are many clients, it can be difficult to find a specific one, so we've added a filtering feature to make searching easier.

chrome-capture-2025-09-19

## Types of changes

## Checklist:

### DIFF
--- a/src/app/stores/clients.ts
+++ b/src/app/stores/clients.ts
@@ -125,7 +125,7 @@ export const useClientsStore = defineStore('Clients', () => {
       };
     });
 
-    // TODO: move search to backend
+    // TODO: move sort to backend
     if (transformedClients !== undefined) {
       transformedClients = sortByProperty(
         transformedClients,


kaaax0815:
I don't think you know all these languages (e.g. the german translation is wrong). only add translations for languages you speak

please move the filtering logic to the backend

note: there was already a similar pr (#1978)

yuWorm:
> I don't think you know all these languages (e.g. the german translation is wrong). only add translations for languages you speak
> 
> 
> 
> please move the filtering logic to the backend
> 
> 
> 
> note: there was already a similar pr (#1978)

Sorry, I didn't check the closed pull request carefully. I looked at that PR, and it's more complete. I'll combine it with the previous PR later, then migrate it to the backend and remove the languages I'm not proficient with.

src/server/api/client/index.get.ts
```diff
@@ -1,6 +1,11 @@
-export default definePermissionEventHandler('clients', 'custom', ({ user }) => {
-  if (user.role === roles.ADMIN) {
-    return WireGuard.getAllClients();
+export default definePermissionEventHandler(
+  'clients',
+  'custom',
+  ({ event, user }) => {
+    const { filter } = getQuery(event);
```
kaaax0815: see other routes on how to use zod

src/app/components/Clients/Search.vue
```diff
@@ -0,0 +1,42 @@
...
+        <IconsClose class="h-3 w-3" />
+      </button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue';
+import { MagnifyingGlassIcon } from '@heroicons/vue/24/outline';
+import IconsClose from '../Icons/Close.vue';
```
kaaax0815: auto import should do its job

src/app/stores/clients.ts
```diff
@@ -31,8 +31,13 @@ export const useClientsStore = defineStore('Clients', () => {
   const clients = ref<null | LocalClient[]>(null);
   const clientsPersist = ref<Record<string, ClientPersist>>({});
 
+  const searchParams = ref({
+    filter: '',
+  });
+
```
kaaax0815, line 37: you could export this directly and avoid the setSearchQuery function call

and just pass that directly as the v-model to the search field

kaaax0815 (changes requested):

src/server/utils/WireGuard.ts
```diff
@@ -134,6 +134,29 @@ class WireGuard {
...
+    }
+
+    if (!filter.trim()) return clients;
+
+    const searchTerm = filter.toLowerCase().trim();
+    return clients?.filter((client) => {
+      const nameMatches = client.name.toLowerCase().includes(searchTerm);
+      const ipv4Matches = client.ipv4Address.toLowerCase().includes(searchTerm);
+      const ipv6Matches = client.ipv6Address.toLowerCase().includes(searchTerm);
+      return nameMatches || ipv4Matches || ipv6Matches;
```
kaaax0815: i think this would make more sense at the db level

kaaax0815 (changes requested):

src/server/utils/WireGuard.ts
yuWorm: ok, I'll revise it and then submit it

src/app/components/Clients/Search.vue
yuWorm: I call setSearchQuery mainly to set clients.value to null. This makes the loading animation show up on the list.

src/app/stores/clients.ts
yuWorm: I call setSearchQuery mainly to set clients.value to null. This makes the loading animation show up on the list.

kaaax0815:
Looks good. I'll try it out in the following days

src/server/database/repositories/client/service.ts
```diff
@@ -42,6 +42,39 @@ function createPreparedStatement(db: DBType) {
         },
       })
       .prepare(),
+    findAllPublicFiltered: db.query.client
+      .findMany({
+        where: or(
+          like(client.name, sql.placeholder('filter')),
+          like(client.ipv4Address, sql.placeholder('filter')),
+          like(client.ipv6Address, sql.placeholder('filter'))
```
kaaax0815, line 50: use `ilike` for case insensitive search. and remove `toLowerCase()` from the inputs

kaaax0815 (changes requested):

src/server/database/repositories/client/service.ts
yuWorm: It looks like sqlite doesn't support ilike

kaaax0815: yes my bad. then client.name should also be lower, otherwise the case-ing would break the search

yuWorm: hmm, it seems like LIKE in SQLite is case-insensitive, but it's case-sensitive in other databases. Since it looks like our project only uses sqlite, can we just ignore the others? I'm thinking of just removing the toLowerCase().

kaaax0815: yeah mb. misread the docs. removing toLowerCase() should be enough

yuWorm:
> Looks good. I'll try it out in the following days

Okay, thanks for your patient help!

kaaax0815:
Thank you very much

PR #2175, opened 2025-09-21
docs: Add AdGuard Home tutorial

ernvk23:
This pull request adds a new, comprehensive tutorial for integrating AdGuard Home with an existing `wg-easy` and Traefik setup.

**Key Features:**
*   Follows the project's multi-compose-file structure for consistency.
*   Provides step-by-step instructions for adding AdGuard Home for network-wide DNS ad-blocking.
*   Uses `wg-easy`'s unattended setup feature for a robust and reproducible configuration.

This guide provides a clear path for users to add DNS filtering to their existing VPN setup.

### DIFF
--- a/docs/content/examples/tutorials/adguard.md
+++ b/docs/content/examples/tutorials/adguard.md
@@ -8,7 +8,13 @@ This tutorial is a follow-up to the official [Traefik tutorial](./traefik.md). I
 
 - A working [wg-easy](./basic-installation.md) and [Traefik](./traefik.md) setup from the previous guides.
 
-## Setup `AdGuardHome`
+/// warning | Important: Following this guide will reset your WireGuard configuration.
+The process involves re-creating the `wg-easy` container and its data, which means **all existing WireGuard clients and settings will be deleted.**
+
+You will need to create your clients again after completing this guide.
+///
+
+## Add `adguard` configuration
 
 1. Create a directory for the configuration files:
 
@@ -39,9 +45,11 @@ services:
             - /etc/docker/volumes/adguard/adguard_conf:/opt/adguardhome/conf
         networks:
             wg:
+                interface_name: eth0
                 ipv4_address: 10.42.42.43
                 ipv6_address: fdcc:ad94:bacf:61a3::2b
-            traefik: {}
+            traefik:
+                interface_name: eth1
         labels:
             - 'traefik.enable=true'
             - 'traefik.http.routers.adguard.rule=Host(`adguard.$example.com$`)'
@@ -64,10 +72,6 @@ Modify the corresponding sections of your existing `wg-easy` compose file to mat
 File: `/etc/docker/containers/wg-easy/docker-compose.yml`
 
 ```yaml
-volumes:
-  etc_wireguard:
-    name: etc_wireguard
-
 services:
   wg-easy:
     ports:
@@ -75,12 +79,13 @@ services:
     ...
     networks:
       wg:
+        interface_name: eth0
         ...
-        interface_name: wgeth
+      traefik:
+        interface_name: eth1
       ...
     ...
     environment:
-      - WG_DEVICE=wgeth
       # Unattended Setup
       - INIT_ENABLED=true
       # Replace $username$ with your username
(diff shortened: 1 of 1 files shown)


docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,159 @@
...
+    networks:
+      traefik:
+        ipv4_address: 10.42.42.43
+        ipv6_address: fdcc:ad94:bacf:61a3::2b
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.adguard.rule=Host(`adguard.$example.com$`)"
+      - "traefik.http.routers.adguard.entrypoints=websecure"
+      - "traefik.http.routers.adguard.service=adguard"
+      - "traefik.http.services.adguard.loadbalancer.server.port=3000"
```
kaaax0815: why not use the default port 80 for adguard here?

docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,159 @@
...
+    ```bash
+    sudo docker network rm traefik
+    ```
+
+    *(It is safe to ignore any error message if the network does not exist.)*
+
+2. Next, create it again with:
+
+    ```bash
+    sudo docker network create --driver=bridge --subnet=10.42.42.0/24 --ipv6 --subnet=fdcc:ad94:bacf:61a3::/64 traefik
```
kaaax0815: wouldn't it make more sense to separate traefik and wg-easy networks? as seen in the traefik example (and include #1795)

ernvk23: @kaaax0815  thanks for the feedback and for the context of issue #1795 
You're right, separating the networks is the better approach for consistency. I'll work on updating the tutorial to reflect the multi-network architecture.
I'll test the changes to ensure everything works and will have the new commit pushed by Friday at the latest.

docs/content/examples/tutorials/adguard.md
ernvk23: @kaaax0815 The decision to use 3000 was to simplify the initial setup for the user.
AdGuard's first-time wizard runs on port 3000, and if not modified, it defaults to using port 80. This ensures the Traefik configuration works immediately without requiring the user to edit the compose file again after the setup is complete.

ernvk23:
@kaaax0815  I've implemented the multi-network architecture as you suggested.
To handle the connection, I've used the traefik.docker.network label. My understanding is that this is the better, native approach for this scenario and it keeps the configuration clean.
I've tested the full wg-easy, traefik, and adguard stack with this new network configuration, and everything works.
Please let me know if this aligns with what you were expecting.

kaaax0815:
I think adguard should join the wireguard network. then static ips should work. and the traefik network should not be modified in any way. just used for traefik nothing else

ernvk23:
@kaaax0815 , thanks for the feedback.
AdGuard is now connected to both networks as you suggested. This approach is definitely cleaner and keeps the existing Traefik setup unmodified.

kaaax0815:
Thank you. Looks good. Can you confirm the latest changes work?

kaaax0815:
Please fix the formatting errors

ernvk23:
@kaaax0815 I need to clarify my testing on the last push.

I had thoroughly tested all previous versions of the PR. For the latest changes, I ran the stack, checked the logs, and confirmed all web endpoints were working. I wrongly assumed the VPN connection would function the same as it had before, but I now realize that was an incorrect assumption and I've hit a networking issue.
I've been debugging it since and have pinpointed the problem:
- DNS queries from a VPN client are reaching AdGuard.
- The DNS reply from AdGuard is not making it back to the client.
This confirms it's a return path failure. I suspect the solution lies with the `WG_POST_UP` and `WG_POST_DOWN` iptables rules, which are likely necessary to handle the NAT in this two-network setup.

I'm working on a fix now.

ernvk23:
@kaaax0815 
I've found a fix. It was indeed an iptables issue.

The root cause is that the `WG_POST_UP` and `WG_POST_DOWN` environment variables are being overwritten by the default Hooks set in the UI.
A temporary solution that works is to manually replace the default rules in the UI's "Hooks" section with the correct ones for this AdGuard setup and then restart the container.
I've updated the docker-compose.yml in the PR to include the correct rules and a comment explaining this workaround.

What are your thoughts on this? I'm wondering if there's a better way to prevent the UI Hooks from overwriting the environment variables.

kaaax0815:
The env vars you described are from v14 and do not work in v15

ernvk23:
@kaaax0815 
I've put everything together and can confirm this now works. The key was setting the iptables rules directly in the UI Hooks. Issue #1795  was also key for getting the multi-network part right. Thanks!

ernvk23:
@kaaax0815 
Pushed a final refinement to the PREROUTING rule for correctness. Tested and works.

docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,172 @@
...
+      - "51820:51820/udp"
+    ...
+    networks:
+      wg:
+        ...
+        interface_name: wgeth
+      ...
+    ...
+    environment:
+      - WG_DEVICE=wgeth
```
kaaax0815: `WG_DEVICE` env var does not exist in v15

docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,172 @@
...
+
+services:
+  wg-easy:
+    ports:
+      - "51820:51820/udp"
+    ...
+    networks:
+      wg:
+        ...
+        interface_name: wgeth
```
kaaax0815: should be `eth0`. as thats the expected device in wg-easy

docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,172 @@
...
+1. Restart `wg-easy`:\
+   _Note: Previous settings and configurations will be restored to default_
+
+    ```shell
+    cd /etc/docker/containers/wg-easy
+    sudo docker compose down -v
+    sudo docker compose up -d
+    ```
+
+2. Edit Wireguard's Hooks.
```
kaaax0815, line 121: there is a template for device. therefore avoid `wgeth` directly. 

are the nat rules to adguard needed?

ernvk23: Got it working with eth0, and now the NAT rules are technically optional, but if used they enforce all DNS queries to be resolved by AdGuard, which prevents leaks, even if clients set their own DNS in the WireGuard config. (Encrypted DNS DoT/DoH remains an exception and will be preferred)
Would you like me to keep/remove them, or mark them as optional?

kaaax0815: you can keep them

ernvk23:
@kaaax0815 
After more testing, I found interface_name: eth0 is unstable for this setup.

In a multi-network container, it causes a race condition leading to intermittent startup failures (Error: file exists). This happens when the system tries to rename the interface to eth0, but that name has already been allocated to the other network.

A stable solution is to use a unique name like wgeth for the interface_name and then target it directly in the POSTROUTING rule.
I've updated the PR with this version.

Let me know what you think.

kaaax0815:
@ernvk23 you have to give every network a name. So if the wg-easy container has network WG and traefik. WG should be eth0 and traefik e.g. eth1

ernvk23:
@kaaax0815 Thanks for pointing me in the right direction, this is definitely better.
Now the race condition is fixed. For consistency, I set the same interface names in AdGuard, even though it's not strictly required. 
This version works as expected.

docs/content/examples/tutorials/adguard.md
```diff
@@ -2,8 +2,172 @@
...
+
+1. Restart `wg-easy`:
+
+    > Note: Previous settings and configurations will be restored to default
+
+    ```shell
+    cd /etc/docker/containers/wg-easy
+    sudo docker compose down -v
+    sudo docker compose up -d
+    ```
```
kaaax0815: either make it a more visible warning (maybe even at the start) or give context on how to add it to a existing installation

ernvk23: Done! Thanks for the suggestion.

kaaax0815 (approved):

ernvk23:
@kaaax0815 
Hi! I’m not sure I fully understand why the type checking is failing, but I think it might not be related to this documentation file specifically. Could you please advise on how to proceed?

kaaax0815:
Typechecking failing is not your fault

Thank you very much

Issue #2174, opened 2025-09-21
[Feat]: Add documentation to allow for routed network instead of NAT


vosdev:
### Description

I want my clients to have a "real" IP on my network and not have them NAT through my WireGuard host's IPv4/IPv6.

There are no steps in the documentation to achieve this. It took me 15 minutes to figure out so I've included the steps in the solution below for others to use. I think this would be a great addition to the documentation. 

It could go under advanced examples.

### Suggested Solution

The below config allows me to achieve this:

docker compose with network_mode host:
```yaml
services:
  wg-easy:
    image: ghcr.io/wg-easy/wg-easy:15
    container_name: wg-easy
    network_mode: "host"
    environment:
      - INSECURE=true #temp
    volumes:
      - ./config:/etc/wireguard              # persistent config
      - /lib/modules:/lib/modules:ro
    cap_add:
      - NET_ADMIN
    devices:
      - /dev/net/tun:/dev/net/tun
    restart: unless-stopped
```

Because we use network_mode host we have to create the sysctl rules on the host itself in `/etc/sysctl.d/90-wireguard.conf`:
```
net.ipv4.ip_forward=1
net.ipv4.conf.all.src_valid_mark=1
net.ipv6.conf.all.disable_ipv6=0
net.ipv6.conf.all.forwarding=1
net.ipv6.conf.default.forwarding=1
```

Either reboot or apply with `sysctl -p /etc/sysctl.d/90-wireguard.conf`

->> In your router, create a static route for both the IPv4 and IPv6 subnet you want your clients to use to your wireguard host

And finally, modify the PostUp and PostDown iptable rules to get rid of the MASQUERADE rules.

PostUp:
```
iptables -A INPUT -p udp -m udp --dport {{port}} -j ACCEPT; iptables -A FORWARD -i wg0 -j ACCEPT; iptables -A FORWARD -o wg0 -j ACCEPT; ip6tables -A INPUT -p udp -m udp --dport {{port}} -j ACCEPT; ip6tables -A FORWARD -i wg0 -j ACCEPT; ip6tables -A FORWARD -o wg0 -j ACCEPT
```

PostDown:
```
[...]

PR #2181, opened 2025-09-22
Add routed (no NAT) example setup

vosdev:
## Description

I've added an example to set up wg-easy with a no-NAT setup.

Fixes https://github.com/wg-easy/wg-easy/issues/2174

## Motivation and Context

No NAT has several benefits!  You can identify your clients by their respective IP addresses instead of seeing all of your clients as your wireguard server's host IP.

## Types of changes

## Checklist:

### DIFF
--- a/docs/content/examples/tutorials/routed.md
+++ b/docs/content/examples/tutorials/routed.md
@@ -2,7 +2,7 @@
 title: Routed setup (No NAT)
 ---
 
-This guide shows how to run **wg-easy** with a routed setup so packets are forwarded instead of NATed.
+This guide shows how to run **wg-easy** with a routed setup, so packets are forwarded instead of NATed.
 
 In a routed design, each WireGuard client keeps its own IPv4/IPv6 address. That means you can identify clients by their real addresses instead of seeing everything as the WireGuard server’s IP.
 
@@ -54,7 +54,7 @@ sysctl -n net.ipv4.ip_forward   # should print 1
 
 ## Add static routes on your router
 
-Pick an IPv4 and IPv6 subnet for your clients and add static routes on your router, pointing to the wireguard server's LAN addresses.
+Pick an IPv4 and IPv6 subnet for your clients and add static routes on your router, pointing to the WireGuard server's LAN addresses.
 
 ### Example
 
@@ -74,11 +74,11 @@ On your router:
 - Route `192.168.0.0/24` → next hop `192.168.10.118`
 - Route `2001:db8:abc:0::/64` → next hop `2001:db8:abc:10:216:3eff:fedb:949e`
 
-Don't forget to create the neccesary firewall rules to allow these subnets to travel across your LAN. Some routers like OPNSense/PFSense may require specific Outbound NAT rules for the chosen IPv4 subnet (And IPv6 if ULA).
+Don't forget to create the necessary firewall rules to allow these subnets to travel across your LAN. Some routers or servers may require specific Outbound NAT rules for the chosen IPv4 and IPv6 subnets to allow traffic to traverse your LAN.
 
-## Wireguard Easy configuration
+## `wg-easy` configuration
 
-In the web UI → Admin → Interface, click Change CIDR and set the IPv4/IPv6 routed subnets you chose above. Save.
+In the Web UI → Admin → Interface, click Change CIDR and set the IPv4/IPv6 routed subnets you chose above. Save.
 
 Then go to Admin → Hooks and add:
 


docs/content/examples/tutorials/routed.md
```diff
@@ -0,0 +1,87 @@
...
+To make use of our own IPv4/IPv6 addresses, run the container with the `network_mode: host` option.
+
+```yaml
+services:
+  wg-easy:
+    image: ghcr.io/wg-easy/wg-easy:15
+    container_name: wg-easy
+    network_mode: "host"
+    environment:
+      - INSECURE=true # TEMPORARY for testing only; remove for production
```
kaaax0815: remove this. best would be to require a existing setup following the existing basic installation

docs/content/examples/tutorials/routed.md
```diff
@@ -0,0 +1,87 @@
...
+
+I want my WireGuard clients in `192.168.0.0/24` and `2001:db8:abc:0::/64`.
+
+- Routed IPv4 subnet: `192.168.0.0/24`
+- Routed IPv6 prefix: `2001:db8:abc:0::/64`
+- WireGuard server IPs: `192.168.10.118` and `2001:db8:abc:10:216:3eff:fedb:949e`
+
+On your router:
+- Route `192.168.0.0/24` → next hop `192.168.10.118`
+- Route `2001:db8:abc:0::/64` → next hop `2001:db8:abc:10:216:3eff:fedb:949e`
```
kaaax0815, line 75: 2001:db8::/32 is a documentation prefix. many people will use the docs as is, therefore either use a correct prefix or add a note

docs/content/examples/tutorials/routed.md
```diff
@@ -0,0 +1,87 @@
...
+Apply and verify:
+```bash
+sysctl -p /etc/sysctl.d/90-wireguard.conf
+sysctl -n net.ipv4.ip_forward   # should print 1
+```
+
+## Add static routes on your router
+
+Pick an IPv4 and IPv6 subnet for your clients and add static routes on your router, pointing to the wireguard server's LAN addresses.
+
```
kaaax0815, line 58: i assume outgoing network works, as the router does the nating. but would happen in a server environment?

vosdev: Either the same (with NAT being done for IPv4 on your router instead of on this wireguard server) or you give your clients public IPv4 addresses and no NAT is being done at all.

kaaax0815: Giving your clients public ipv4 addresses is possibly insecure and very expensive.

Either way I'll try out the setup and report back

vosdev: > Giving your clients public ipv4 addresses is possibly insecure and very expensive.

I'm aware, we did this for years at my previous job. But we were swimming in public IPv4 addresses and simply did not wish to do NAT anywhere in our network at all. Obviously no inbound traffic was allowed to these IPs.

Anyone who has access to large numbers of IPv4 public addresses to give your VPN clients public IPv4 addresses should have plenty of in-house knowledge to add a basic firewall in front of their network. I don't think it's something we need to worry about.

docs/content/examples/tutorials/routed.md
vosdev: > 2001:db8::/32 is a documentation prefix.

That's why I used it :). It is the correct prefix to use. I will add a note

docs/content/examples/tutorials/routed.md
vosdev: I can add a warning that this guide / type of set-up is for more advanced users if that has your preference

kaaax0815: I do like the no nat solution. because with pihole and stuff many people complain that it won't show the ip of the individual client. this would fix that.

best would be a more general solution

kaaax0815:
There is this discussion. https://github.com/wg-easy/wg-easy/discussions/1168

In an server environment you would have to enable nat for outbound traffic. and the command in the answer should do it.  (at least thats how i understand it)
Maybe it makes sense to add that?

kaaax0815:
Please fix the formatting errors

vosdev:
> There is this discussion. #1168
> 
> In an server environment you would have to enable nat for outbound traffic. and the command in the answer should do it. (at least thats how i understand it) Maybe it makes sense to add that?

That differs per router software/manufacturer. Most do it by default, the advanced ones do not. Those who use a more advanced advanced type of router software/manufacturer have the knowledge to get this done. I have added a note that it might be required. We can't include step by step instructions for every router manufacturer that exists on the planet. They have their own documentation for that :)

vosdev:
> Please fix the formatting errors

I'd love to but the job only says that there are formatting errors. Not which ones or on what line.

kaaax0815:
Formatting looks good.

Would you rephrase the not about opnsense or add a note about outbound nat rules.
and if you assign a ULA (spelling error in docs) to the clients, wouldn't you need rules for ipv6 too not only for ipv4?

docs/content/examples/tutorials/routed.md
```diff
@@ -0,0 +1,95 @@
...
+- Routed IPv4 subnet: `192.168.0.0/24`
+- Routed IPv6 prefix: `2001:db8:abc:0::/64`
+- WireGuard server IPs: `192.168.10.118` and `2001:db8:abc:10:216:3eff:fedb:949e`
+
+On your router:
+
+- Route `192.168.0.0/24` → next hop `192.168.10.118`
+- Route `2001:db8:abc:0::/64` → next hop `2001:db8:abc:10:216:3eff:fedb:949e`
+
+Don't forget to create the neccesary firewall rules to allow these subnets to travel across your LAN. Some routers like OPNSense/PFSense may require specific Outbound NAT rules for the chosen IPv4 subnet (And IPv6 if ULA).
```
kaaax0815: I think something like "Some routers (or servers) may require specific Outbound NAT rules for the chosen IPv4 and IPv6 subnet" would make more sense and applies to most cases

kaaax0815 (changes requested):

docs/content/examples/tutorials/routed.md
kaaax0815: @vosdev any update?

kaaax0815 (approved):

vosdev:
@kaaax0815 thanks for the commit. I was unable to due to personal health issues.

Thanks for the merge!

PR #2182, opened 2025-09-23
Fixed oneTimeLink filename encoding

hongry18:
# Fix 500 error when downloading files with non-ASCII filenames

## Description
Fixed a 500 error that occurred when downloading configuration files with non-ASCII characters (e.g., Korean, Chinese, Japanese) in the client name. The error was caused by invalid characters in the Content-Disposition header when the client name contained non-ASCII characters.

**Changes made:**
- Applied proper URL encoding to `client.name` in the Content-Disposition header using `encodeURIComponent()`
- This ensures that non-ASCII characters are properly encoded according to RFC standards

## Motivation and Context
**Problem:** When a client has a non-ASCII name (like Korean "사용자"), downloading the configuration file resulted in a 500 error with the message:
```
Error: Invalid character in header content ["Content-Disposition"]
```

**Root cause:** HTTP headers must contain only ASCII characters. The Content-Disposition header was trying to include raw non-ASCII characters, which is not allowed by HTTP specifications.

**Solution:** Encode the filename using `encodeURIComponent()` to ensure all characters are properly escaped for HTTP headers.

## How has this been tested?
- **Test environment:** Local development server
- **Test case 1:** Client with ASCII name (e.g., "user") - ✅ Works as before
- **Test case 2:** Client with Korean name (e.g., "사용자") - ✅ Now downloads successfully without 500 error
- **Test case 3:** Client with other non-ASCII characters - ✅ Should work with any Unicode characters

**Before fix:**
```javascript
client.name = "사용자"
// Content-Disposition: attachment; filename="사용자.conf"
// Result: 500 Error - Invalid character in header content
```

**After fix:**
```javascript
client.name = "사용자"
// Content-Disposition: attachment; filename="%EC%82%AC%EC%9A%A9%EC%9E%90.conf"
// Result: File downloads successfully
```

N/A - This is a backend fix for HTTP header encoding.

## Types of changes

## Checklist:

### DIFF
--- a/src/server/routes/cnf/[oneTimeLink].ts
+++ b/src/server/routes/cnf/[oneTimeLink].ts
@@ -30,7 +30,7 @@ export default defineEventHandler(async (event) => {
   setHeader(
     event,
     'Content-Disposition',
-    `attachment; filename="${client.name}.conf"`
+    `attachment; filename="${encodeURIComponent(client.name)}.conf"`
   );
   setHeader(event, 'Content-Type', 'text/plain');
   return config;


kaaax0815:
I think it would make more sense to use the existing cleaning logic from here https://github.com/wg-easy/wg-easy/blob/9b29d72/src/server/api/client/%5BclientId%5D/configuration.get.ts

PR #2226, opened 2025-10-20
AmneziaWG 2.0

cany748:
## Description
This PR adds support for AWG 2.0.
solves https://github.com/wg-easy/wg-easy/issues/2053

## Types of changes

## Checklist:

### DIFF
--- a/src/server/utils/types.ts
+++ b/src/server/utils/types.ts
@@ -27,7 +27,7 @@ export const MtuSchema = z
   .max(9000, { message: t('zod.mtu') });
 
 export const JcSchema = z.number().min(1).max(128).nullable();
-// TODO: more validation
+
 export const JminSchema = z.number().max(1279).nullable();
 
 export const JmaxSchema = z.number().max(1280).nullable();


src/server/utils/types.ts
```diff
@@ -25,6 +25,18 @@ export const MtuSchema = z
   .min(1280, { message: t('zod.mtu') })
   .max(9000, { message: t('zod.mtu') });
 
+export const JcSchema = z.number().min(1).max(128).nullable();
+// TODO: more validation
```
cany748: Checking these parameters is complicated because they depend on the MTU and other parameters.
Do we need precise checking here, or can we leave it as is?

https://github.com/amnezia-vpn/amneziawg-linux-kernel-module#configuration

src/app/pages/admin/config.vue
```diff
@@ -47,6 +47,16 @@
...
       </FormGroup>
+      <FormGroup v-if="globalStore.information?.isAwg">
+        <FormHeading>Amnezia</FormHeading>
+        <FormNumberField id="jC" v-model="data.defaultJC" label="Jc" />
+        <FormNumberField id="jMin" v-model="data.defaultJMin" label="Jmin" />
+        <FormNumberField id="jMax" v-model="data.defaultJMax" label="Jmax" />
+        <FormTextField id="i1" v-model="data.defaultI1" label="I1" />
+        <FormTextField id="i2" v-model="data.defaultI2" label="I2" />
+        <FormTextField id="i3" v-model="data.defaultI3" label="I3" />
+        <FormTextField id="i4" v-model="data.defaultI4" label="I4" />
```
cany748: The templates contain TS errors due to a conflict between undefined and null types.

kaaax0815:
looks good. seems to be not a breaking change.

there should be descriptions for all the fields.

one thing i thought about is already existing installations. as e.g. h1 needs a computed default it will not be used as the migration will use the default for integer which is 0. this should be noted in the docs

and about defaults. are the params something users would want to change? as there is no fallback to the user config, this would mean a user has to change every client

cany748:
> there should be descriptions for all the fields.

complete

> one thing i thought about is already existing installations. as e.g. h1 needs a computed default it will not be used as the migration will use the default for integer which is 0. this should be noted in the docs

I changed the logic for generating random values ​​for H1-H4

> and about defaults. are the params something users would want to change? as there is no fallback to the user config, this would mean a user has to change every client

the user should be able to disable the fields he needs. If you clear all the fields associated with awg, you'll get a regular wireguard.

cany748:
Additional checks can be added to the zod schema using `refine`, as well as checking of `I` fields using regex. I'm not sure yet if this is necessary in this PR.

The PR is almost completely ready. There's only one remaining issue: a type conflict in the interface (null and undefined).

src/app/pages/admin/interface.vue
```diff
@@ -21,6 +21,109 @@
...
+          :description="$t('awg.s1Description')"
+        />
+        <FormNumberField
+          id="s2"
+          v-model="data.s2"
+          :label="$t('awg.s2Label')"
+          :description="$t('awg.s2Description')"
+        />
+
+        <div class="col-span-full text-sm">* {{ $t('awg.*') }}</div>
```
kaaax0815: avoid `*` be more descriptive

src/app/pages/admin/interface.vue
```diff
@@ -21,6 +21,109 @@
           :description="$t('admin.interface.deviceDesc')"
         />
       </FormGroup>
+      <FormGroup v-if="globalStore.information?.isAwg">
+        <FormHeading>AmneziaWG Obfuscation Parameters</FormHeading>
```
kaaax0815: translate too (also see admin config and client config)

src/server/utils/WireGuard.ts
```diff
@@ -5,6 +5,9 @@ import type { InterfaceType } from '#db/repositories/interface/types';
 
 const WG_DEBUG = debug('WireGuard');
 
+const generateRandomHeaderValue = () =>
```
kaaax0815, line 8: maybe switching to `node:crypto` would make sense?

cany748: It doesn't make sense. These numbers are not involved in encryption.

src/server/utils/types.ts
```diff
@@ -25,6 +25,18 @@ export const MtuSchema = z
...
+export const JminSchema = z.number().max(1279).nullable();
+
+export const JmaxSchema = z.number().max(1280).nullable();
+
+export const SSchema = z.number().max(1132).nullable();
+
+export const HSchema = z.number().min(5).max(2147483647).nullable();
+
+export const ISchema = z.string().nullable();
```
kaaax0815, line 39: is there a reason for making these all nullable? if everything is null why not just use wg instead of awg?

kaaax0815:
> Additional checks can be added to the zod schema using refine, as well as checking of I fields using regex. I'm not sure yet if this is necessary in this PR.

I don't think so. We should provide some examples in the docs tho

src/server/utils/types.ts
cany748: All fields must be nullable. Not all of them are always needed. We must give the user the freedom to flexibly customize

issafram:
I was overseas very recently and my Wireguard (wg-easy) connection wouldn't work.  Likely due to the ISP using DPI to detect the Wireguard handshake.  I'm hoping this gets merged in and potentially fixes that issue.

kaaax0815:
@cany748 if you have any questions or need help feel free to ask

cany748:
> if you have any questions or need help feel free to ask

pr is ready for merge.
I suggest adding documentation in a separate pr

kaaax0815:
Thank you very much

issafram:
I don't have a build on my PC, so I'll just ask.  
Does this allow us to specify if we want to use the base WireGuard vs Amnezia?

I'm considering having 2 containers.  One for normal WireGuard and another for Amnezia.

cany748:
> Does this allow us to specify if we want to use the base WireGuard vs Amnezia?

yes

kaaax0815:
@cany748 As this seems to work just fine. Are you willing to write some docs (even if only some bullet points or examples) so we can release v15.2.0

PR #2228, opened 2025-10-21
Fix: Allow lower MTU

### LINKED ISSUE #2210: [Bug]: Auto MTU seems not to work
### Describe the bug

Not sure if a bug but anyway.
I set up a fresh instance (v15.1.0) with default the Android client MTU is set to `auto` (placeholder text). I can connect to the server, i get `Latest Handshakes`, i can ping hosts in my home nework and i can curl HTTP traffic. I cannot curl HTTPS traffic, the response is `unexpected eof while reading`.
If, however, set the MTU on the client manually to 1150, it works perfectly. Since i'm not sure whether this is acutally a good idea and this does not scale, to do that manually on each client, i don't want to do that.
Server side i cannot set the MTU below 1280. ChatGPT recommended following hook: `iptables -t mangle -A FORWARD -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu` but apperently this is not supported.
On v14 this worked out of the box without any MTU configurations.
If this matters: running in Kubernetes with Cilium cni in native routing mode, wireguard port bound to loadbalancerip populated via Metallb IP. If any more info is needed, let me know.

### Expected behavior

Work just fine!

### Relevant log output

```shell

```

kaaax0815:
## Description

For users only using IPv4 a lower MTU might be required. This PR allows that. But users using IPv6 should never go below 1280

## Motivation and Context

Closes #2210

### DIFF
--- a/src/server/utils/types.ts
+++ b/src/server/utils/types.ts
@@ -22,7 +22,8 @@ export const EnabledSchema = z.boolean({ message: t('zod.enabled') });
 
 export const MtuSchema = z
   .number({ message: t('zod.mtu') })
-  .min(1280, { message: t('zod.mtu') })
+  // min for IPv6 is 1280, but we allow lower for IPv4
+  .min(1024, { message: t('zod.mtu') })
   .max(9000, { message: t('zod.mtu') });
 
 export const PortSchema = z


theklausklausen:
This for sure fixes my problem. Copilot wrote it changes the min MTU from 1280 to 1024 but removes it entirely. Not sure if this actually has any implications but I would recommend to keep a minimum.

PR #2229, opened 2025-10-22
add INIT_PASSWORD_FILE environment variable

a-skuba:
## Description
This PR adds the `INIT_PASSWORD_FILE` environment variable to read the password from a file during unattended setup. The `INIT_PASSWORD_FILE` variable takes precedence over the `INIT_PASSWORD` variable.

## Motivation and Context

Using secrets in Docker setup is preferred and more secure way of passing secrets to a runtime.

## How has this been tested?

I have tested, that it reads provided file and fallback to `INIT_PASSWORD`. If read fails, it will throw an exception.

## Types of changes

## Checklist:

### DIFF
--- a/docs/content/advanced/config/unattended-setup.md
+++ b/docs/content/advanced/config/unattended-setup.md
@@ -10,7 +10,7 @@ These will only be used during the first start of the container. After that, the
 | ------------------ | ---------------------------- | --------------------------------------------------------- | ----- |
 | `INIT_ENABLED`     | `true`                       | Enables the below env vars                                | 0     |
 | `INIT_USERNAME`    | `admin`                      | Sets admin username                                       | 1     |
-| `INIT_PASSWORD`    | `Se!ureP%ssw`                | Sets admin password                                       | 1     |
+| `INIT_PASSWORD`    | `Se!ureP%ssw`                | Sets admin password (or `INIT_PASSWORD_FILE`)             | 1     |
 | `INIT_HOST`        | `vpn.example.com`            | Host clients will connect to                              | 1     |
 | `INIT_PORT`        | `51820`                      | Port clients will connect to and wireguard will listen on | 1     |
 | `INIT_DNS`         | `1.1.1.1,8.8.8.8`            | Sets global dns setting                                   | 2     |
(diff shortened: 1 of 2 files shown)


src/server/utils/config.ts
```diff
@@ -1,5 +1,6 @@
 import debug from 'debug';
 import packageJson from '@@/package.json';
+import fs from 'fs';
```
kaaax0815, line 3: its best practice to use the `node:` protocol

src/server/utils/config.ts
```diff
@@ -31,10 +32,14 @@ export const WG_ENV = {
   EXPERIMENTAL_AWG: process.env.EXPERIMENTAL_AWG === 'true',
 };
 
+const PASSWORD = process.env.INIT_PASSWORD_FILE?.length > 0 ?
```
kaaax0815, line 35: use a simple truthy check for `INIT_PASSWORD_FILE`. length > 0 does not add anything.
the error should be handled (e.g. if file does not exist)

docs/content/advanced/config/unattended-setup.md
```diff
@@ -10,7 +10,7 @@ These will only be used during the first start of the container. After that, the
 | ------------------ | ---------------------------- | --------------------------------------------------------- | ----- |
 | `INIT_ENABLED`     | `true`                       | Enables the below env vars                                | 0     |
 | `INIT_USERNAME`    | `admin`                      | Sets admin username                                       | 1     |
-| `INIT_PASSWORD`    | `Se!ureP%ssw`                | Sets admin password                                       | 1     |
+| `INIT_PASSWORD`    | `Se!ureP%ssw`                | Sets admin password (or `INIT_PASSWORD_FILE`)             | 1     |
```
kaaax0815, line 13: this should be documented better

src/server/utils/config.ts
a-skuba: If the file does not exists and the variable is set, I would expect container to fail, especially as this is initializing and a security configuration. If we gracefully handled the exception, what should be the password? Let me know what you think.

kaaax0815: i don't think it should be handled gracefully. but with a proper error message

PR #2267, opened 2025-11-06
Add wgHelper tests

egmen:
Prepare for refactor wgHelpers to improve future functionality

## Description
Added tests for one file and test stage to github ci

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -66,3 +66,31 @@ jobs:
           cd src
           pnpm install
           pnpm ${{ matrix.command }}
+
+  test:
+    name: Test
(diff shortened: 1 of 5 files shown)


kaaax0815:
What about using nuxt test utils. https://nuxt.com/docs/3.x/getting-started/testing

egmen:
> What about using nuxt test utils. https://nuxt.com/docs/3.x/getting-started/testing

A never used `@nuxt/test-utils` so can't add it wisely.
But it uses `vitest`, so I think adding it in the future wouldn't be a problem

kaaax0815:
should be recreated after #2226

Issue #1225, opened 2024-07-28
Download config file as a .conf instead of .txt


altbert:
Will it be possible to change the default download extension of the client configuration file with the extension ".conf" instead of ."txt".
I propose this change to avoid the step of renaming the file when trying to import it from the WireGuard app on Android or iPhone.

tetuaoro:
It's already with `.conf` extension. I tested with the android, ios and laptop.

altbert:
I see, from Firefox PC the file download as .conf but from Firefox mobile as .txt.
Seems that Firefox mobile on Android changes the extension to .txt.
Probably a Firefox Behavior.
![wg-admin](https://github.com/user-attachments/assets/68e76ec8-aff5-43ad-98dc-43eb597fe4a9)

eandersons:
See also #417 and #796 for more information regarding this.

Issue #796, opened 2024-01-29
Download configuration file as .conf file


navilg:
**Describe the bug**
Currently when we download configuration file of a peer, It downloads as .txt file. But to import the configuration using wireguard app, It needs either .conf or .zip file.

**To Reproduce**

**Expected behavior**
When we download a configuration file, It should be downloaded as .conf file which is compatible extension for wireguard.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: WIndows, Linux, Android and iOS

**Smartphone (please complete the following information):**

**Additional context**

eandersons:
Similar issue: #417 

This seems to be a browser specific problem, but since you did not care to provide detailed information, it is hard to tell if this is the case.

navilg:
Oh. I thought it application issue and cared not to share browser details. Here are more details

Desktop (please complete the following information):

    OS: Ubuntu 23.04
    Browser [e.g. chrome, safari]: Firefox
    Version [e.g. 22]: 122.0

Smartphone (please complete the following information):

    Device: [e.g. iPhone6]: Samsung
    OS: [e.g. iOS8.1]: Android 14
    Browser [e.g. stock browser, safari]: Firefox
    Version [e.g. 22]: 122.0.1

eandersons:
Yes, this seems to be the same issue as in #417.

For me on Arch Linux peer configuration file is downloaded with correct file name with both Firefox (122.0) and Firefox Developer Edition (123.0b5), but not with Fennec (122.0.0; based on Firefox for Android) on Android phone.

As I mentioned in #417:

> Downloaded file's name is determined from the header [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) which [WireGuard Easy sends](https://github.com/wg-easy/wg-easy/blob/910ce7052e5e50d9d6c814f8fea6e1a01e7fb96b/src/lib/Server.js#L122) (for example: `Content-Disposition: attachment; filename="client-name.conf"`), but seems like Firefox on mobile overwrites the extension [..]

WireGuard Easy sets the header `Content-Disposition` with `peer-name.conf` in it:

https://github.com/wg-easy/wg-easy/blob/910ce7052e5e50d9d6c814f8fea6e1a01e7fb96b/src/lib/Server.js#L122

Response in Fennec also contains `peer-name.conf`:

```
content-disposition: attachment; filename="peer-name.conf"
```

[...]

Issue #417, opened 2022-11-15
Client Configuration file format is wrong


XxAcielxX:
Hello,

I'm using your wg-easy docker container, when I click the download button to get the clients' config, the file format comes with a `.txt`. Which is not the format accepted by WireGuard Software/App. The correct file format should be `.conf`.

Kindly look into to this.

eandersons:
For me client configuration file is downloaded with the extension `.conf`. At least in Firefox (Developer Edition) and Chromium.

Additional information (like operating system, browser, possibly `docker run` command or `docker-compose.yml` that was used to create the WireGuard Easy container, reverse-proxy in use (if any) and its configuration) could help others to possibly recreate the issue and spot the problem that is causing to download client configuration with the extension `.txt` instead of `.config`.

XxAcielxX:
Using Firefox on Android Phone.
![Screenshot_20221115-151259~2](https://user-images.githubusercontent.com/21063358/201893945-1e2c1472-7115-4bb5-a2b4-83328a4c0f5a.jpg)


docker-compose
```
...
  wg-easy:
    image: weejewel/wg-easy
    container_name: wg-easy
    cap_add:
      - NET_ADMIN
      - SYS_MODULE
    sysctls:
      - net.ipv4.ip_forward=1
      - net.ipv4.conf.all.src_valid_mark=1
    networks:
      adnio:
        ipv4_address: ${SUBNET}${WG_EASY_IP}
    ports:
      - ${WG_PORT}:51820/udp
    environment:
      - PASSWORD=${WG_EASY_PASSWD}
      - WG_HOST=${WG_SUBDOMAIN}
      - WG_ALLOWED_IPS=${WG_ALLOWED}
      - WG_DEFAULT_ADDRESS=${WG_SUBNET}
      - WG_PORT=${WG_PORT}
      - WG_DEFAULT_DNS=${SUBNET}${PIHOLE_IP}
    volumes:
      - ${APP_CONF}/wg-easy:/etc/wireguard
    restart: unless-stopped
```

Using LSIO SWAG as Reverse-Proxy. And I access wg-easy as `https://wg.mydomain.ltd`

eandersons:
It seems that this issue is specific to mobile Firefox.
On my phone in Bromite (Chromium based) configuration is downloaded with the suffix `.conf`, but on Fennec (Firefox based) `.txt` is used.

[...]

PR #2269, opened 2025-11-09
Fixed a bug causing .conf.txt download on Android affecting Firefox based Android browsers.

bedro0:
## Motivation and Context
The extra step of renaming the config file causes mild annoyance.
Issues: [#1225](https://github.com/wg-easy/wg-easy/issues/1225), [#796](https://github.com/wg-easy/wg-easy/issues/796), [#417](https://github.com/wg-easy/wg-easy/issues/417).

Seems like the browser is interpreting the content header and renaming the file extension accordingly. Changing it to Octet Stream should force the browser to treat it as a generic file instead of a text file.

## How has this been tested?
Tested the config download feature on multiple devices, including Desktop, iOS and Android (Chromium and Firefox on each platform). Since the only thing changed is a header format, this should not affect any other elements of the app.

## Screenshots:
Before:
Screenshot_20251109-022830
After:
Screenshot_20251109-022935

## Types of changes

## Checklist:

### DIFF
--- a/src/server/routes/cnf/[oneTimeLink].ts
+++ b/src/server/routes/cnf/[oneTimeLink].ts
@@ -32,6 +32,6 @@ export default defineEventHandler(async (event) => {
     'Content-Disposition',
     `attachment; filename="${WireGuard.cleanClientFilename(client.name) || client.id}.conf"`
   );
-  setHeader(event, 'Content-Type', 'text/plain');
+  setHeader(event, 'Content-Type', 'application/octet-stream');
   return config;
 });


kaaax0815:
Should be added here https://github.com/wg-easy/wg-easy/blob/master/src/server/routes/cnf/%5BoneTimeLink%5D.ts too probably

issafram:
Thanks for this.  I ran into this issue a week or so ago

kaaax0815:
Thank you very much

bedro0:
Thanks for merging, and thanks for catching the other instance of this. I was not looking for it, nor was I aware of its existence. :)

PR #2295, opened 2025-11-21
fix(i18n): Add missing translation for delete action

ChiahongHong:
## Description
This PR adds a missing translation key for the delete action used in the client details page.  The hard-coded `Delete` label in the Vue component is replaced with a localized string reference.

Additionally, several new Traditional Chinese (zh-TW) translation entries are added, including items related to configuration viewing and copy-status messages.

## Motivation and Context
The UI previously displayed an untranslated `Delete` string when rendered in non-English locales.

## How has this been tested?
- Verified that the client details page now correctly shows the localized delete label.

## Types of changes

## Checklist:

### DIFF
--- a/src/app/pages/clients/[id].vue
+++ b/src/app/pages/clients/[id].vue
@@ -179,7 +179,7 @@
               @delete="deleteClient"
             >
               <FormSecondaryActionField
-                label="Delete"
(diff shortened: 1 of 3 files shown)


kaaax0815:
Good find. Thank you very much

PR #2339, opened 2025-12-11
feat: add automatic public IP refresh for dynamic IP users

SurceBeats:
## Summary

-   Adds optional automatic public IP refresh feature for users with dynamic IPs
-   When enabled, periodically checks if public IP has changed and updates the database
-   Fully backwards compatible, disabled by default

## Problem

When using wg-easy with `SERVERURL=auto`, the public IP is detected once during initial setup and stored in the database (`user_configs_table.host`). If the ISP assigns a new IP address, new client configurations are generated with the stale IP, making them unusable.

Current workarounds require manual database updates, external scripts with Docker command overrides or reinstalling wg-easy.

## Solution

Two new environment variables:

  ------------------------------------------------------------------------------------
  Variable                            Default             Description
  ----------------------------------- ------------------- ----------------------------
  `SERVERURL_AUTO_REFRESH`            `false`             Enable automatic IP refresh
  `SERVERURL_AUTO_REFRESH_INTERVAL`   `300`               Seconds between IP checks
  ------------------------------------------------------------------------------------

### How it works

1.  The existing `cronJob()` (runs every 60s) checks if auto refresh is enabled
2.  If enabled and the configured interval has passed, fetches current public IP via OpenDNS (same existing methodd)
3.  Compares it with the stored IP in the database
4.  If different, updates `user_configs_table.host`
5.  New client configs and QR codes automatically use the updated IP

### Usage

``` yaml
services:
  wg-easy:
    environment:
      - SERVERURL_AUTO_REFRESH=true
      - SERVERURL_AUTO_REFRESH_INTERVAL=300
```

## Changes

  ----------------------------------------------------------------------------
  File                              Changes
  --------------------------------- ------------------------------------------
  `src/server/utils/config.ts`      Added WG_AUTO_REFRESH_ENV configuration
  `src/server/utils/ip.ts`          Exported `getCurrentPublicIpv4()` function
  `src/server/utils/WireGuard.ts`   Added `checkAndUpdatePublicIp()` method and integration with `cronJob()`
  ----------------------------------------------------------------------------

## Test plan

-   Verify feature is disabled by default
-   Enable `SERVERURL_AUTO_REFRESH=true` and verify IP check every 5 min
-   Test with custom interval
-   Verify logs when IP changes: `Auto-refresh: IP changed from X to Y`
-   Verify new client configs use updated IP after change
-   Test behavior when IP detection fails (warning and continue)

### DIFF
--- a/src/server/utils/WireGuard.ts
+++ b/src/server/utils/WireGuard.ts
@@ -1,6 +1,8 @@
 import fs from 'node:fs/promises';
 import debug from 'debug';
 import { encodeQR } from 'qr';
+import { getCurrentPublicIpv4 } from './ip';
+import { WG_AUTO_REFRESH_ENV } from './config';
 import type { InterfaceType } from '#db/repositories/interface/types';
 
 const WG_DEBUG = debug('WireGuard');
@@ -9,6 +11,8 @@ const generateRandomHeaderValue = () =>
   Math.floor(Math.random() * 2147483642) + 5;
 
 class WireGuard {
+  private lastIpCheck = 0;
+
   /**
    * Save and sync config
    */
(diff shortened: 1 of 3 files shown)


kaaax0815:
i think a ddns service makes more sense for this

SurceBeats:
The ddns is exactly the reason why this pull requests exists, once IP changes nothing works.

Here's the flow explaining why DDNS doesn't solve this issue, let’s say:

1. You have DDNS configured → `myvpn.duckdns.org` points to your IP
2. You configure wg-easy with `SERVERURL=auto`
3. wg-easy detects your public IP (e.g., `85.123.45.67`) and **stores it in the database**
4. Your ISP changes your IP to `85.123.45.99`
5. DuckDNS updates correctly → `myvpn.duckdns.org` → `85.123.45.99` ✓
6. **But wg-easy keeps generating configs with `85.123.45.67`** because it's hardcoded in the DB

kaaax0815:
what? you can simply set the Host in the Admin panel to `myvpn.duckdns.org`. and wireguard will resolve the correct ip, wg-easy doesnt do any of that

i don't know where you got the `SERVERURL` env var from

SurceBeats:
Never mind, I get it now, sorry, somehow I ended up mixing up env vars from linuxserver/wireguard. Setting the DDNS hostname directly in the admin panel makes sense, anyway I can confirm this Frankenstein PR worked detecting and changing the IP lol

PR #2375, opened 2026-01-03
Fix healthcheck for amneziawg

SKProCH:
## Description
Fixes healthcheck for awg

## Motivation and Context
Currently healthcheck for awg doesn't work, since wg command return nothing when awg interface is up

Works for me

## Types of changes

## Checklist:


kaaax0815:
Please create an issue for this. I think the logic with env vars and such should be handled in the healthcheck. and using a http route for that makes most sense imo

Issue #2376, opened 2026-01-03
[Bug]: Healthcheck not working for amneziawg


SKProCH:
### Describe the bug

When using the amneziawg, healthcheck always fail due to `wg show` doesn't list awg interfaces resulting in docker container shows as unhealthy

### Expected behavior

Healthcheck should work properly with amneziawg

### Relevant log output

```shell

```

The-Engineer:
Same here, my container restarts every 3 minutes (version 15.2.1) running on ubuntu server but no logs are provided. Monitoring the container confirms that failing health checks are the reason

rushiiMachine:
You can change the healthcheck to use the `awg` binary rather than the `wg` binary. This is currently hardcoded at the moment in the Dockerfile, but if you want to change it in a docker compose setup:
```yml
services:
    wireguard:
        image: ghcr.io/wg-easy/wg-easy:15.2.1
        # ...
        healthcheck:
          test: /usr/bin/timeout 5s /bin/sh -c "/usr/bin/awg show | /bin/grep -q interface || exit 1"
          interval: 1m
          timeout: 5s
          retries: 3
```

The-Engineer:
thanks @rushiiMachine that works. Could a solution for the image be to move healthchecks to a healthcheck.sh and put old and new command in there with a switch to choose via env var? The image helathcheck then just calls the bash script

rushiiMachine:
> with a switch to choose via env var

I assume it's not that simple, since the default "auto" behavior when `OVERRIDE_AUTO_AWG` is unset depends on whether the AmneziaWG kernel module is available.

The-Engineer:
I proposed a change for the heathcheck handling the technology chosen automatically but it was closed by @kaaax0815 without discussion or real reason sandly. I will use that approach for now as it solves all problems I currently have and makes the amnezia aproach work.

experiment-everything:
@The-Engineer are you basically using AWG as the main wireguard instance, or integrated into default wireguard and using AWG as a separate service? Heard and read good things about AWG over default, but I just learned about Easy WG so still reading up

PR #2409, opened 2026-01-18
feat: add Amnezia check to Dockerfile healthcheck

The-Engineer:
## Description
This extends the Docker health check to check first if the WG interface is present and in a second step checks if the AWG interface is present. If none are found the check fails.
The '2>/dev/null' silences errors for the case that the AWG kernel module is not installed.

## Motivation and Context
When using the image with Amnezia my contianer constantly restarts. In Issue [2376](https://github.com/wg-easy/wg-easy/issues/2376) I got the recommendation on how to fix this temporarilly by adding a different healthcheck to my compose file.
This change adds this into the image directly.

## How has this been tested?
I wrote this check in compose files on different systems running:
- wireguard with amnezia kernel module
- wireguard without amnezia kernel module
- amnezia

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,8 @@ RUN apk add linux-headers build-base git && \
 FROM docker.io/library/node:jod-alpine
 WORKDIR /app
 
-HEALTHCHECK --interval=1m --timeout=5s --retries=3 CMD /usr/bin/timeout 5s /bin/sh -c "/usr/bin/wg show | /bin/grep -q interface || exit 1"
+HEALTHCHECK --interval=1m --timeout=5s --retries=3 \
+  CMD /usr/bin/timeout 5s sh -c '/usr/bin/wg show 2>/dev/null | grep -q interface || /usr/bin/awg show 2>/dev/null | grep -q interface'
 
 # Copy build
 COPY --from=build /app/.output /app


kaaax0815:
duplicate of https://github.com/wg-easy/wg-easy/pull/2375

The-Engineer:
May I kindly ask for a discussion here or in the issue regarding this open problem? Without a change to the healthcheck the image constantly restarts right now when using Amnezia.
The other mentions PR tries to solche the same [issue](https://github.com/wg-easy/wg-easy/issues/2376) but with a less resilient approach

kaaax0815:
The other pr is almost the exact same. And I mentioned the reason for closing the pr there, which applies to this pr too

The-Engineer:
The other PR does not handle errors if you want to use Amnezia but do not have the kernel module installed. As far as I understand you prefer a approach which selects healthchecks via env vars but that would not be a very stable approach as it does not handle user error like the missing kernel module and makes the solution more complex

kaaax0815:
> does not handle user error like the missing kernel module

it should match the wg-easy behavior otherwise it does not make much sense

if you select awg via env vars and don't have the kernel module, wg-easy will fail, and the healthcheck should too

The-Engineer:
But the proposed approach matches the wg-easy behaviour and does not need changing when env var names change or user behaviour is not on point. From neither a users perspektive nor a dev perspective I see an advantage in using the env 
Why bind to env vars when there is an automatic solution possible?

PR #2238, opened 2025-10-27
feat: use iptables filter traffics,only forward to allowip and port.

laoshancun:
Enhance Security by Restricting Traffic Forwarding to Specified AllowIPs and Ports

## Description

This PR significantly improves the security posture of the application by implementing the following changes:

1. Removed the default behavior that forwards all traffic through the VPN tunnel
2. Implemented traffic restriction logic to only forward traffic related to the specified AllowIPs
3. Added support for port-specific traffic restrictions within individual AllowIP entries
4. Refactored IP processing functions by extracting and reusing common utilities

The implementation includes:
- Moving the `isIPv6` and `parseIpAndPort` functions to a shared utility module
- Updating client traffic restriction plugin to use these shared utilities
- Modifying client configuration generation to properly handle IP addresses with port specifications
- Implementing iptables/ip6tables rules that restrict traffic to only the explicitly allowed IPs and ports
image

image

## Motivation and Context

The primary motivation for this change is to enhance security by implementing the principle of least privilege. Previously, the default configuration forwarded all traffic through the VPN, which could potentially expose internal networks or sensitive services.

Key security benefits include:

1. **Granular Access Control**: Users can now precisely specify which IPs and ports should be accessible through the VPN
2. **Reduced Attack Surface**: Limiting traffic to only necessary destinations decreases potential attack vectors
3. **Better Compliance**: The ability to restrict traffic aligns with common security compliance requirements
4. **Improved Network Segmentation**: Administrators can more effectively segment network access based on client needs

This change addresses a fundamental security concern while maintaining flexibility through the AllowIPs configuration, which now supports port-level granularity.

## How has this been tested?

The security enhancements have been thoroughly tested through multiple verification methods:

1. **Functional Testing**: Verified that traffic is only forwarded to explicitly allowed IPs and ports
2. **Security Validation**: Confirmed that unauthorized traffic destinations are properly blocked
3. **Regression Testing**: Ensured that existing VPN functionality remains intact for authorized connections
4. **Edge Case Testing**: Tested various IP formats (IPv4, IPv6, CIDR notation) with and without port specifications

All tests were conducted in both development and production-like environments, confirming the security improvements work as expected without introducing new issues.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring (non-breaking change that improves code quality and maintainability)

[...]

Issue #1375, opened 2024-09-08
[Feat]: Hint when wg-easy server's IP is not included in AllowedIPs


lars-devs:
### Description

Hi, first kudos to this awesome project! ❤
I recently noticed that limiting WG traffic to communication between clients (we built a ZeroTier-like alternative) make the connection stats on the wg-easy page disappear, since even when connected to the WG server, it doesn't receive information from the clients.

I changed the preset config line 
`- WG_ALLOWED_IPS=10.0.0.0/24`
to
`- WG_ALLOWED_IPS=10.0.0.0/24, x.x.x.x/32`
to make it appear back again.

### Suggested Solution

I'd suggest a hint, just like the update notification, if `WG_ALLOWED_IPS` is set to something else than `0.0.0.0` and `WG_HOST`'s IP isn't included inside the list of allowed IP's.

![grafik](https://github.com/user-attachments/assets/eed13da9-fff6-4cd5-bcbb-d50a573e4cc6)

### Alternatives

To allow to disable the hint, in case the configuration is intended, an additional flag to suppress the hint would be possible.

kaaax0815:
But not including the IP of the WG Server is supported afaik. And editing is for advanced users (which should know what they are doing)

lars-devs:
Hi, thanks for the reply. I fully agree on that! Not including the server's IP is totally fine since on usual wireguard installations, there's no need to do that, but it directly affects a feature of weg-easy (connection status) when not doing so. 

That's why I suggested the hint as a suggestion without forcing more advanced users to run their server as they like. In that case the flag to suppress the hint would be a neat solution to satisfy as many needs as possible, imo. What do you think?

kaaax0815:
is the connection status still shown (if peer is connected)?. Or does even this not work

lars-devs:
Connection stats are only shown if AllowedIPs is set to 0.0.0.0 or the server's IP is included. Otherwise, e.g. 10.0.0.0/24 or with the server's IP included, there's no visible difference between connected and not connected clients. Basically, they appear as offline. Communication using wireguard is working fine.

kaaax0815:
Thanks. Weird that even the connection status doesn't work (as its connected to wireguard anyway)

kaaax0815:
We could also just add `10.8.0.1/32` (if WG_DEFAULT_ADDRESS is the default) to the allowed ips. As this would not break anything and makes everything work
@lars-devs Whats your opinion

[...]

PR #2418, opened 2026-01-24
Feature/client firewall filtering

lanrat:
Implement server-side firewall rules to restrict client network access,
allowing administrators to enforce security policies that cannot be
bypassed by clients modifying their local configuration.

This feature addresses the limitation where "Allowed IPs" only controls
client-side routing but doesn't prevent clients from accessing networks
they shouldn't reach. The firewall rules are enforced on the server
using iptables/ip6tables and provide true access control.

Features:
- Opt-in via "Enable Per-Client Firewall" toggle in admin interface
- Per-client "Firewall Allowed IPs" field for granular control
- Support for IPs, CIDRs, and port-based filtering
- Protocol specification: TCP, UDP, or both (default)
- IPv4 and IPv6 dual-stack support
- Falls back to client's allowedIps when firewallIps is empty
- Clean separation of routing (allowedIps) from security (firewallIps)

Supported formats:
- 10.10.0.3 (single IP)
- 10.10.0.0/24 (CIDR range)
- 192.168.1.5:443 (IP with port, both TCP+UDP)
- 192.168.1.5:443/tcp (IP with specific protocol)
- `[2001:db8::1]:443` (IPv6 with port)

Implementation:
- New database columns: firewall_enabled (interfaces), firewall_ips (clients)
- Migration 0003_add_firewall_filtering for schema updates
- firewall.ts utility for iptables chain management (WG_CLIENTS chain)
- Integration into WireGuard.ts for automatic rule application
- UI components with conditional rendering based on firewall toggle

Technical details:
- Uses custom WG_CLIENTS iptables chain for isolation
- Rebuild strategy: flush and recreate all rules on config save
- Mutex protection via rebuildInProgress/rebuildQueued flags
- Graceful cleanup when firewall is disabled
- No new dependencies (uses existing is-ip, is-cidr packages)

## How has this been tested?

This has been tested extensibility with docker-compose with multiple clients and various firewall rules.
Edge cases and race-conditions have also been tested, as well as input validation.

image

image

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -122,7 +122,7 @@
     "config": "Configuration",
     "viewConfig": "View Configuration",
     "firewallIps": "Firewall Allowed IPs",
-    "firewallIpsDesc": "Destination IPs/CIDRs this client can access (server-side enforcement). Leave empty to use Allowed IPs. Supports port and protocol filtering: 10.0.0.1:443, 10.0.0.0/24:443/tcp, [::1]:443/udp"
+    "firewallIpsDesc": "Destination IPs/CIDRs this client can access (server-side enforcement). Leave empty to use Allowed IPs. Supports optional port and protocol filtering. See docs for syntax."
   },
   "dialog": {
     "change": "Change",
@@ -203,7 +203,7 @@
       "address6": "IPv6 Address",
       "serverAllowedIps": "Server Allowed IPs",
       "firewallIps": "Firewall Allowed IPs",
-      "firewallIpsInvalid": "Invalid firewall IP entry. Must be: IP (10.0.0.1), CIDR (10.0.0.0/24), IP:port (10.0.0.1:443), CIDR:port (10.0.0.0/24:443), IP:port/protocol (10.0.0.1:443/tcp), or IPv6 ([::1], [::1]:443, [::1]:443/udp)"
+      "firewallIpsInvalid": "Invalid firewall IP entry. See docs for supported syntax."
     },
     "user": {
       "username": "Username",


lanrat:
This offers a more robust implementation of the iptables filtering proposed in #2238

kaaax0815:
Looks quite good. But the problem with this is that we don't have a firewall abstraction layer (which would be a lot of work). But that means anyone wanting to use nftables or host networking can't use this feature

lanrat:
I understand the benefits for a possible firewall abstraction layer, and I agree on a pure nftables system, this feature would not currently work.

However if the host supports iptables, I don't see why host networking would not work. Additionally this feature is disabled by default. I could add a check and throw an error or prevent the user from enabling this feature if the system does not support it.

If nftables support was added, would you be open to merging?

kaaax0815:
I am open to merging this regardless. But it should be documented clearly

dl-team-ops:
Thanks for this PR, this is very useful for enforcing server-side filtering.
We have opened a separate issue to discuss a related requirement around routing intent vs client configuration.
https://github.com/wg-easy/wg-easy/issues/2436

lanrat:
@kaaax0815 I've just updated the PR to include updated documentation about the iptables dependency. 

I also updated the UI so if the admin attempts to enable the firewall feature on a system without iptables, they will get an informative error message.

I tested all of these features.

Let me know if you would like any other changes.

src/server/utils/types.ts
```diff
@@ -1,6 +1,8 @@
 import type { ZodSchema } from 'zod';
 import z from 'zod';
 import type { H3Event, EventHandlerRequest } from 'h3';
+import { isIP } from 'is-ip';
+import { default as isCidr } from 'is-cidr';
```
kaaax0815: importing the default like that makes no sense

src/server/utils/WireGuard.ts
```diff
@@ -2,6 +2,7 @@ import fs from 'node:fs/promises';
 import debug from 'debug';
 import { encodeQR } from 'qr';
 import type { InterfaceType } from '#db/repositories/interface/types';
+import { firewall } from './firewall';
```
kaaax0815: auto import should already import it

src/nuxt.config.ts
```diff
@@ -131,6 +126,7 @@ export default defineNuxtConfig({
     },
     alias: {
       '#db': fileURLToPath(new URL('./server/database/', import.meta.url)),
+      '#utils': fileURLToPath(new URL('./server/utils/', import.meta.url)),
```
kaaax0815: the utils folder is already auto imported through nuxt

src/server/utils/firewall.ts
```diff
@@ -0,0 +1,294 @@
+import debug from 'debug';
+import { isIPv6 } from 'is-ip';
+
+import { exec } from './cmd';
```
kaaax0815: should be auto imported

docs/content/guides/clients.md
```diff
@@ -19,7 +19,35 @@ Which IPs will be routed through the VPN.
 
 This will not prevent the user from modifying it locally and accessing IP ranges that they should not be able to access.
 
-Use firewall rules to prevent access to IP ranges that the user should not be able to access.
+Use the Firewall Allowed IPs feature to prevent access to IP ranges that the user should not be able to access.
+
+## Firewall Allowed IPs
+
+!!! note
+    This field only appears when **Per-Client Firewall** is enabled in the Admin Panel → Interface settings.
```
kaaax0815: this is not the correct syntax for a note

docs/content/guides/clients.md
```diff
@@ -19,7 +19,35 @@ Which IPs will be routed through the VPN.
...
+
+Server-side firewall rules that restrict which destinations the client can access, regardless of their local configuration.
+
+Unlike "Allowed IPs" which only controls routing on the client side, these rules are enforced by the server using iptables/ip6tables and cannot be bypassed by the client.
+
+**Supported Formats:**
+
+- `10.10.0.3` - Allow access to a single IP address
+- `10.10.0.0/24` - Allow access to an entire subnet
```
kaaax0815: what about cidr + port. or cidr + port + proto. or ip + proto. or cidr + proto

kaaax0815:
We should make the feature as experimental for now. as having a hole in the firewall could be pretty bad.
we should add docs so the user knows how to disable the firewall feature on a per client basis (e.g. by adding 0.0.0.0/0, ::/0) or similar

src/server/utils/firewall.ts
lanrat: As far as I can tell Nuxt does not support auto-importing 3rd party packages.

src/server/utils/types.ts
lanrat: This should be fixed.

src/server/utils/WireGuard.ts
lanrat: Fixed.

src/nuxt.config.ts
lanrat: fixed.

docs/content/guides/clients.md
lanrat: A proto requires a port, the rest should be supported.

docs/content/guides/clients.md
lanrat: fixed.

docs/content/guides/clients.md
lanrat: The documentation has been updated to better show what is and is not supported, and a check on save has also been added.

lanrat:
@kaaax0815 I made all of the requested changes/updates/notes. PTAL.

src/server/utils/firewall.ts
kaaax0815: i am talking about this line `import { exec } from './cmd';`

src/nuxt.config.ts
kaaax0815: the utils folder is still added as an alias even tho it doesnt need to be

src/server/utils/firewall.ts
lanrat: I missed that one. Fixed in the latest commit.

src/nuxt.config.ts
lanrat: I missed that one. Fixed in the latest commit.

lanrat:
@kaaax0815 is there anything else you want to see in the PR to get it merged?

src/nuxt.config.ts
```diff
@@ -139,5 +134,6 @@ export default defineNuxtConfig({
   alias: {
     // for typecheck reasons (https://github.com/nuxt/cli/issues/323)
     '#db': fileURLToPath(new URL('./server/database/', import.meta.url)),
+    '#utils': fileURLToPath(new URL('./server/utils/', import.meta.url)),
```
kaaax0815: files from utils are auto imported

kaaax0815: github still showed me the change even tho you already fixed it...

kaaax0815:
I think adding unit tests make sense here. as the parsing is quite complex. i'll look into it. you don't have to do it

kaaax0815:
I added vitest and rebased your pr. please create some unit tests

kaaax0815:
I added some unit tests. It seems that simple ipv6 addresses fail. there are some typescript errors that should be fixed.
feel free to simplify the unit tests

lanrat:
I've updated and fixed the tests, fixed the typescript and lint errors, and also added comments to the firewall rules.

kaaax0815:
please recreate your migrations

lanrat:
Migrations recreated, and rebased.

kaaax0815:
See https://github.com/wg-easy/wg-easy/issues/1375. I think we should add a note to the docs that the ip of wg-easy should be included

lanrat:
I added a note to the docs.

src/i18n/locales/en.json
```diff
@@ -196,7 +201,9 @@
       "expiresAt": "Expires At",
       "address4": "IPv4 Address",
       "address6": "IPv6 Address",
-      "serverAllowedIps": "Server Allowed IPs"
+      "serverAllowedIps": "Server Allowed IPs",
+      "firewallIps": "Firewall Allowed IPs",
+      "firewallIpsInvalid": "Invalid firewall IP entry. Must be: IP (10.0.0.1), CIDR (10.0.0.0/24), IP:port (10.0.0.1:443), or IP:port/protocol (10.0.0.1:443/tcp)"
```
kaaax0815: can also be cidr with port or cidr with port and protocol

lanrat: Done.

lanrat:
> i think we should properly mark the feature as experimental for now

@kaaax0815 what do you mean by this? The firewall is already disabled by default and is opt-in.

We can update the documentation and admin page to indicate this feature is experimental if you would like.

I'm not a fan of putting it behind a general experimental flag as that removes the ability for admins to have granular control on which "experimental" features they want enabled vs disabled.

kaaax0815:
> We can update the documentation and admin page to indicate this feature is experimental if you would like.

I think that makes sense

> I'm not a fan of putting it behind a general experimental flag as that removes the ability for admins to have granular control on which "experimental" features they want enabled vs disabled.

Agreed, as the feature is completely opt in, there is no reason for putting it behind a flag

shurkanTwo:
I am also highly interested in this feature

src/i18n/locales/en.json
```diff
@@ -203,7 +203,7 @@
       "address6": "IPv6 Address",
       "serverAllowedIps": "Server Allowed IPs",
       "firewallIps": "Firewall Allowed IPs",
-      "firewallIpsInvalid": "Invalid firewall IP entry. Must be: IP (10.0.0.1), CIDR (10.0.0.0/24), IP:port (10.0.0.1:443), or IP:port/protocol (10.0.0.1:443/tcp)"
+      "firewallIpsInvalid": "Invalid firewall IP entry. Must be: IP (10.0.0.1), CIDR (10.0.0.0/24), IP:port (10.0.0.1:443), CIDR:port (10.0.0.0/24:443), IP:port/protocol (10.0.0.1:443/tcp), or IPv6 ([::1], [::1]:443, [::1]:443/udp)"
```
kaaax0815: technically you forgot ipv6 cidr. but maybe it makes more sense to refer to the documentation?

lanrat: I can add it if you want, but I don't think its the job for every error message and tooltip to have the full documentation.

docs/content/guides/admin.md
```diff
@@ -2,4 +2,42 @@
...
-TODO
+## Interface Settings
+
+### Per-Client Firewall
+
+Enable server-side firewall filtering to enforce network access restrictions per client.
+
+When enabled, each client can have custom "Firewall Allowed IPs" configured that restrict which destinations they can access through the VPN. These restrictions are enforced by the server using iptables/ip6tables and cannot be bypassed by the client.
+
+/// warning | Experimental Feature
```
lanrat, line 13: @kaaax0815 We already have a warning in the documentation that this is experimental.

lanrat:
@kaaax0815 do you feel the existing experimental feature warning is sufficient to get this merged?

src/i18n/locales/en.json
kaaax0815: yes, thats we should properly just refer to the docs, instead of adding some formats, that could become invalid in the future

lanrat: In that case, is this safe to resolve?

werter-l:
+1
Add the ability to add addresses in bulk. That would be very convenient!

kaaax0815:
> Add the ability to add addresses in bulk. That would be very convenient!

should be handled in a separate pr.

P.S: avoid +1, just add a reaction to the pr

src/i18n/locales/en.json
kaaax0815: i think you should remove the examples and just refer to the docs. this would avoid invalidations, and properly explain all possible formats.
something like "pls refer to the docs for valid entries/formats"

lanrat: Done: https://github.com/wg-easy/wg-easy/pull/2418/commits/18cca02f6090e4a6c8cae973b64ec0d32a2e2660

kaaax0815:
Thank you very much

kaaax0815:
the links in the docs are bad, need to be fixed

Issue #2417, opened 2026-01-23
[Feat]: QR code generation for AmneziaVPN app


ne0x:
### Description

The project already supports generating QR codes for the WireGuard app.
However, AmneziaVPN uses a different QR format and currently cannot be configured via QR code generated by wg-easy.

Many users rely on AmneziaVPN, and adding native QR support would improve the onboarding and configuration experience for these users, while keeping the existing WireGuard QR functionality unchanged.

I have prepared a UI mockup that demonstrates how this feature could be integrated into the existing QR modal without breaking current behavior. You can try this generated demo QR🙂
![Image](https://github.com/user-attachments/assets/cb4e2c49-9e30-4844-826e-c95e487b155f)

### Suggested Solution

Extend the existing QR code modal with an optional AmneziaVPN mode. This was briefly discussed here:
https://github.com/wg-easy/wg-easy/issues/2053#issuecomment-3790471594

UI:
- Add a small toolbar in the bottom-left corner of the QR modal
- The toolbar is shown only when the current environment supports an AmneziaWG configuration
- By default, the modal displays the standard WireGuard QR code
- A toggle (or “A” icon) allows switching between:
  - WireGuard QR
  - AmneziaVPN-compatible QR

Behavior:
- When switching to AmneziaVPN mode:
  - the client requests an AmneziaVPN-formatted QR code from the server
  - the QR code updates dynamically
- When switching back:
  - the standard WireGuard QR is shown again

Notes:
- The “Download” button shown in the mockup is optional and included only as a possible future extension
- I have already implemented and tested AmneziaVPN QR generation in a separate project: https://github.com/ne0x/awg-easy
- I’m ready to implement this feature here and prepare a PR if the approach and UI direction are acceptable

Questions:
- The AmneziaVPN QR format includes a `description` field. After importing the configuration into AmneziaVPN, this value is used as the server title.
  What would you recommend storing in this field?
  In my implementation, I used `client.name`, but I’m open to better suggestions.
- Feedback on the proposed UI approach would be very welcome, as UI/UX design is not my primary focus.

ne0x:
almost done

<img width="1280" height="742" alt="Image" src="https://github.com/user-attachments/assets/41c4bd1b-b168-4605-bd67-99b5a4a19a1a" />

cany748:
[...]

PR #2423, opened 2026-01-26
Feat: QR code generation for AmneziaVPN app #2417

ne0x:
https://github.com/wg-easy/wg-easy/issues/2417

## Description
Adds an AmneziaVPN QR option to the client QR modal and backend. The UI toggle switches between standard WireGuard QR and AmneziaVPN-compatible QR. The API now accepts `type=amnezia-vpn` on `/api/client/:id/qrcode.svg` to return an AmneziaVPN-specific QR payload. Also adds i18n strings for the toggle labels.

## Motivation and Context
AmneziaVPN uses a different QR format, so the existing WireGuard QR cannot be imported. This change adds a compatible QR payload while preserving the current default WireGuard QR behavior. The feature is available regardless of `EXPERIMENTAL_AWG` and `OVERRIDE_AUTO_AWG`, and generates a compatible payload for both `wg` and `awg` server modes.

## How has this been tested?
- Tested QR import in AmneziaVPN on two servers:
  - `wg` mode
  - `awg` mode
- Verified the default QR (no query param) still generates the standard WireGuard QR and remains unchanged.

## Screenshots:
AmneziaVpnQR
  
awg_app

## Types of changes

## Checklist:


src/server/utils/wgHelper.ts
```diff
@@ -164,6 +165,112 @@ PersistentKeepalive = ${client.persistentKeepalive}
...
+    const [dns1 = '', dns2 = ''] = (client.dns ?? userConfig.defaultDns ?? []).filter(
+      (value) => /^\d{1,3}(?:\.\d{1,3}){3}$/.test(value)
+    );
+
+    let awgExtras: Record<string, string> = {};
+    let containerType = 'wireguard';
+
+    if (wgExecutable === 'awg') {
+      containerType = 'awg';
+      // S3, S4, i1, i2, i3, i4, i5 are not supported by AmneziaVPN app for now
```
cany748: This is not true! The AmneziaVPN app supports these parameters. Configurations without these parameters are called AmneziaWG Legacy.

https://docs.amnezia.org/documentation/instructions/new-amneziawg-selfhosted/

cany748: It is important to note that using these QR codes for AWG is highly discouraged. Using AWG without i1-i5 parameters currently fails to do its main job, which is to bypass DPI.

src/server/database/repositories/client/types.ts
```diff
@@ -103,3 +103,12 @@ export type ClientCreateFromExistingType = Pick<
   | 'publicKey'
   | 'enabled'
 >;
+
+const qrType = z.preprocess(
+  (val) => (val === 'amnezia-vpn' ? val : undefined),
+  z.enum(['amnezia-vpn']).optional()
+);
```
kaaax0815: a fallback to `wireguard` instead of undefined makes more sense

src/server/utils/wgHelper.ts
```diff
@@ -1,3 +1,4 @@
+import { deflateSync as zlibDeflateSync } from 'node:zlib';
```
kaaax0815: use async await instead of sync

src/app/components/Clients/QRCodeDialog.vue
```diff
@@ -18,4 +29,6 @@
 
 <script setup lang="ts">
 defineProps<{ qrCode: string }>();
+
+const useAmnezia = ref(false);
```
kaaax0815, line 33: shouldn't this be dependent on the use of awg? does it make sense to use the amneziavpn app for normal wireguard?

ne0x: For example, I already have AmneziaVPN on my device, and it supports standard WireGuard and adds backward-compatible traffic obfuscation. I use it.

src/server/utils/wgHelper.ts
ne0x: I noticed that the AmneziaVPN app excludes these parameters from the current section when I add a TXT config.
I simply mirrored the same logic used when adding a TXT config.

In any case, these fields are still present in `last_config.config`, where I store the configuration in its original text form.
They may still be used by the app internally, but this needs to be verified.

src/server/database/repositories/client/types.ts
ne0x: This was intentional, so that the original route continues to work by default without any parameters.
That’s why I simply drop the parameter if it’s not `type=amnezia-vpn`.

If you think it makes more sense to handle this explicitly (e.g. fallback to `wireguard`), I’m fine with changing it.

kaaax0815: using a fallback would keep backwards compatibility, but it makes it clearer internally

src/server/utils/wgHelper.ts
ne0x: Ok. I’ll update this accordingly to avoid the sync call.
Something like this:
```
import { deflate } from 'node:zlib';
import { promisify } from 'node:util';

const zlibDeflate = promisify(deflate);
const compressed = await zlibDeflate(plain);
```

src/app/components/Clients/QRCodeDialog.vue
kaaax0815: true, i just don't like advocating for a third party product, where the main purpose is not even being a simple vpn app

src/server/utils/wgHelper.ts
ne0x: We're waiting. AmneziaVPN 4.8.12.7 is expected to be released with support for AmneziaWG 2.0 protocol.

src/app/components/Clients/QRCodeDialog.vue
ne0x: I understand. The decision is yours.

src/server/database/repositories/client/types.ts
```diff
@@ -104,11 +104,10 @@ export type ClientCreateFromExistingType = Pick<
   | 'enabled'
 >;
 
-const qrType = z.preprocess(
-  (val) => (val === 'amnezia-vpn' ? val : undefined),
-  z.enum(['amnezia-vpn']).optional()
-);
+const qrType = z.enum(['amnezia-vpn', 'wireguard']).catch('wireguard');
```
kaaax0815: why not use `.optional().default('wireguard')`

kaaax0815:
Please remove the ai generated translations

src/app/components/Clients/QRCodeDialog.vue
kaaax0815: i don't like the current design with amneziavpn as a toggle. i would consider a tabbed interface. or a collapsible with the a name like Third Party Clients or Unofficial Clients

kaaax0815:
The question still remains if the support for AmneziaVPN is worth it

ne0x:
> The question still remains if the support for AmneziaVPN is worth it

That’s a fair question.

From my side, the motivation was mainly to support an existing workflow where users already rely on AmneziaVPN for importing configurations (via `.conf` files or QR codes), and to keep feature parity with that experience.

That said, I completely understand if supporting AmneziaVPN is considered out of scope for this project.
If that’s the case, I’m fine with maintaining this functionality in my own fork for my specific needs.

Happy to adapt or drop the AmneziaVPN-related parts depending on what makes the most sense for the project.

kaaax0815:
i think it makes sense for now to e.g. make a request for the amneziavpn app to support generic wireguard configs.

if they don't, we can see again if the feature makes sense

PR #2433, opened 2026-01-26
Add clients pagination

### LINKED ISSUE #2441: [Bug]: Webui gets very slow with many clients (> 200)
### Describe the bug

Hi,

We use wg-easy to manage around 215 VPN clients. As the version 0.15 show realtime stats, the webui starts to get very slow (almost frozen).

Would be nice to have an option to disable realtime stats on the UI (disabling charts isn't enough to solve the issue).

Tested with Firefox and Chromium.

### Expected behavior

Work just fine!

### Relevant log output

```shell

```

dew1ad:
## Description
This PR adds pagination and backend sorting support to the clients list.

The backend API was extended to support paginated and sorting queries, including
proper total pages calculation. The frontend was updated to consume the new API
and display pagination controls.

## Motivation and Context
Previously, the clients list loaded all records at once, which caused unnecessary
database load and did not scale well with a growing number of clients.

This change improves performance, scalability, and usability by loading only
the required subset of clients and allowing users to navigate between pages
and sort results.

## How has this been tested?
- Manually tested pagination with different page and limit values
- Verified filtering combined with pagination
- Checked sorting behavior (ASC/DESC)
- Validated API query parameters
- Confirmed frontend pagination controls work as expected

image

## Types of changes

## Checklist:


src/app/components/Clients/Pagination.vue
```diff
@@ -0,0 +1,33 @@
+<template>
```
kaaax0815, line 1: use `radix-vue` for the pagination component

kaaax0815 (changes requested):

src/app/components/Clients/Pagination.vue
dew1ad: Fixed
image

dew1ad: light theme
image

kaaax0815:
The queries can be simplified and by making the filter nullable you could also avoid some

dew1ad:
> The queries can be simplified and by making the filter nullable you could also avoid some

I’m not entirely sure what you mean by simplifying the queries.
On the backend side, the new query parameters are optional (can be undefined),
so they are ignored when not provided.

Could you please point me to a specific file or query where you think this
simplification would be most appropriate? That would help me understand
the expected approach better.

kaaax0815:
I think it should be possible to refactor `findPaginatedListPublicFiltered` and `findPaginatedListPublic` into one. and their respective count queries. this would clean up the filtering code aswell (as this introduced the duplication)

kaaax0815:
Closing as stale

PR #2438, opened 2026-01-27
Update Dockerfile to make better use of cached resources

lanrat:
## Description

I've reordered some of the steps in the dockerfiles so that system packages do not need to be re-installed on re-builds if only the application source code changes. I also added cache mounts which will allow dependencies to be cached between builds.

Overall this brought Docker builds on my system down from 4-5 minutes. to 2 minutes.

## Motivation and Context

When developing, waiting 4-5 minutes between builds was frustrating when a lot of the time was spent re-installing the same dependencies.

## How has this been tested?

* verified docker images build
* verified the app starts and all features continue to work
* Timed the builds when making changes to the source code and noticed a 2-3 minute improvement.

## Types of changes

This is a quality of life improvement. I'm not sure it fits any of the above options.

## Checklist:

- [X] My code follows the code style of this project.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,42 +6,26 @@ RUN npm install --global corepack@latest
 # Install pnpm
 RUN corepack enable pnpm
 
+# Build amneziawg-tools
+ARG AMNEZIAWG_VERSION=v1.0.20250903
+RUN apk add linux-headers build-base git && \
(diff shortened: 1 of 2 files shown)


kaaax0815:
`pnpm build` without any cache takes around 1,5m on my system.
`pnpm dev` takes around 25s to start

Anyway. Caching the system install is bad imo, as then they won't be updated until their cache is busted. and i don't know when that will happen

PR #2453, opened 2026-02-03
fix: correct help text for prometheus metrics

### LINKED ISSUE #2382: [Bug]: Invalid Metric Name in HELP Line of Prometheus Exposition
### Describe the bug

When attempting to scrape metrics from **wg-easy** using **Telegraf** (specifically the `inputs.prometheus` plugin), the following error occurs:

`"decoding response failed: text format parsing error in line 1: invalid metric name in comment"`

After reviewing the **OpenMetrics** specification (which the Prometheus text format follows), it appears that the metrics generated by **wg-easy** violate the protocol's grammar. This does not seems to be a bug in Telegraf's parser, but rather an issue with how the `HELP` metadata is being formatted in wg-easy.

**The following line currently produced by wg-easy is invalid:**

```text
# HELP wg-easy and wireguard metrics
```

### **Why this is invalid (OpenMetrics Specification)**

According to the OpenMetrics ABNF grammar:

1. **Metric Name Character Set:** A `metricname` must only contain letters, numbers, or underscores (`ALPHA / DIGIT / "_"`). It **cannot** contain hyphens (`-`).
2. **Mandatory Association:** The `HELP` keyword must be followed by a valid `metricname`, a space, and then the docstring.
3. **The Problem:** In the example `# HELP wg-easy and wireguard metrics`, the parser identifies `wg-easy` as the metric name. Because it contains a **hyphen**, the parser rejects the entire stream as malformed.

### Expected behavior

The metrics endpoint should strictly follow the OpenMetrics/Prometheus exposition format. If a `HELP` line is provided, it must reference a valid metric name (using underscores) or be formatted as a standard comment (starting with `# ` but without the `HELP` or `TYPE` keywords) to be ignored by the parser.

**The expected output should follow this format:**

```text
# HELP wireguard metrics

# HELP wireguard_configured_peers
# TYPE wireguard_configured_peers gauge
wireguard_configured_peers{interface="wg0"} 8
```

### Relevant log output

```shell

```

thebigpotatoe:
## Description

Fixes #2382. 

## Motivation and Context

I tried deploying this recently and came across the same issue in my setup. Seems like this should be a quick and easy fix. I'm no TS expert or knowledgable on the repo, but I think I found the right place to change this.

## How has this been tested?

Haven't

## Types of changes

## Checklist:

### DIFF
--- a/src/server/routes/metrics/prometheus.get.ts
+++ b/src/server/routes/metrics/prometheus.get.ts
@@ -39,8 +39,6 @@ async function getPrometheusResponse() {
   const id = `interface="${wgInterface.name}"`;
 
   const returnText = [
-    '# HELP wg-easy and wireguard metrics',
-    '',
     '# HELP wireguard_configured_peers',
     '# TYPE wireguard_configured_peers gauge',
     `wireguard_configured_peers{${id}} ${wireguardPeerCount}`,


src/server/routes/metrics/prometheus.get.ts
```diff
@@ -39,8 +39,6 @@ async function getPrometheusResponse() {
   const id = `interface="${wgInterface.name}"`;
 
   const returnText = [
-    '# HELP wg-easy and wireguard metrics',
```
kaaax0815, line 42: i think changing `wg-easy` to `wg_easy` or removing `wg-easy` makes more sense than removing the whole line

thebigpotatoe: I actually had that first, second guessed myself and asked chat for help who said that it's not compliant. Silly idea.

I've just read [the open metrics spec](https://prometheus.io/docs/specs/om/open_metrics_spec/) itself and it seems to suggest making it an underscore should be ok, but it needs to follow the HELP specification the same as the other metrics as there is no header specification. 

Looks like it has to be;

```
# HELP target Target metadata
```

So it would be better to have something like the following;

```
# HELP wg_easy wg_easy Metrics for wg_easy
```

Or delete it as it currently is committed as again there is no header specification for the spec. 

Let me know which way you'd prefer, happy to make the change. 

Thanks for a great tool!

kaaax0815: i think you are correct. help is used to describe a field but here it describes the entire response, which makes no sense as the user knows about the entrypoint anyway if they use it

kaaax0815:
Thank you very much

PR #2553, opened 2026-03-18
Small code quality improvements

Anghios:
- Fix misleading JSDoc comment in cache.ts
- Mitigate timing-based username enumeration in Basic auth
- Extract duplicated TOTP configuration into private method
- Replace manual peer counter with clients.length in Prometheus metrics
- Simplify isValidPasswordHash return expression

### DIFF
--- a/src/server/database/repositories/user/service.ts
+++ b/src/server/database/repositories/user/service.ts
@@ -58,6 +58,17 @@ export class UserService {
     this.#statements = createPreparedStatement(db);
   }
 
+  #createTotp(user: Pick<UserType, 'username' | 'totpKey'>) {
+    return new TOTP({
+      issuer: 'wg-easy',
+      label: user.username,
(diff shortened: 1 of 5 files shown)


src/server/database/repositories/user/service.ts
```diff
@@ -58,6 +58,17 @@ export class UserService {
...
   }
 
+  #createTotp(user: Pick<UserType, 'username' | 'totpKey'>) {
+    return new TOTP({
+      issuer: 'wg-easy',
+      label: user.username,
+      algorithm: 'SHA1',
+      digits: 6,
+      period: 30,
+      secret: user.totpKey!,
```
kaaax0815: avoid the non null assertion. instead make the totpkey in the parameter required

kaaax0815 (changes requested):

kaaax0815 (approved):

PR #2556, opened 2026-03-21
Create a client using existing public key

### LINKED ISSUE #2235: [Feat]: Ability to manually add peer public key
### Description

Current behavior is whenever we add a peer, a public-private key pair is created and stored in the server while allowing us to re-download it again and again. But in fact server doesn't need to store peer private keys once they are acquired by the peer. 

### Suggested Solution

I guess there are two options:

- Allowing admin to add public keys of new peers. In this case the private keys won't be stored at the server since they were created outside of it.
- Removing config download option. In this case the config file only can be downloaded right after the creation of the peer.

### Alternatives

_No response_

### Additional Context

I am not a security specialist but just thought that since the private keys are not needed for the server to operate, there is no point to store them in there while clients have already got them

Dvorhack:
## Description

I added the possibility to provide a wireguard public key during the client creation so that the server doesn't have the private key of the user

I added a new optional field in the new user form and if this field is provided, the backend doesn't create a new keypair.
In the case where the user provide a public key, the private key is placeholded using REDACTED. Then the user must change the generated config to insert his private key 

## Motivation and Context

I have a use case where my users creates their own keys pairs and only give me the public key

## How has this been tested?

I tested the default setup, using the default compose.yml

image

## Types of changes

## Checklist:


kaaax0815:
PR looks good. Please fix lint errors

PR #2594, opened 2026-04-28
fix(ui):Error pops up when enabling disabled client

### LINKED ISSUE #2584: [Bug]: Expired client can be re-enabled via toggle but gets silently disabled again
### Describe the bug

**Description**
When a client has an expiration date set and that date has passed, the toggle in the UI allows you to re-enable the client. However, the client gets automatically disabled again shortly after with no warning, error message, or log entry explaining why.

**Current behavior**
The toggle appears to work, the client shows as enabled briefly, then silently flips back to disabled. No log output is generated. The interval between enabling and auto-disabling is inconsistent — it can happen almost instantly or take up to ~30 seconds, making it harder to diagnose.

**Steps to reproduce**

Create a client and set an expiration date in the past
Client gets automatically disabled (correct behavior)
Toggle the client back on via the UI
Client appears enabled briefly then gets disabled again with no feedback (inconsistent behavior)

**Version**
wg-easy v15.2

### Expected behavior

Either (for me the best one):
Prevent re-enabling an expired client via the toggle and show a clear message ("This client is expired, please update the expiration date first")
Or:
automatically clear the expiration date when manually re-enabling a client

### Relevant log output

```shell

```

FelipeCordova:
Action is prevented and a clear message is displayed in Web UI

Fixes #2584

### DIFF
--- a/src/server/api/client/[clientId]/enable.post.ts
+++ b/src/server/api/client/[clientId]/enable.post.ts
@@ -12,7 +12,11 @@ export default definePermissionEventHandler(
     const client = await Database.clients.get(clientId);
     checkPermissions(client);
 
-    if (client && client.expiresAt !== null && new Date() > new Date(client.expiresAt)) {
+    if (
+      client &&
+      client.expiresAt !== null &&
+      new Date() > new Date(client.expiresAt)
+    ) {
       throw createError({
         statusCode: 422,
         statusMessage:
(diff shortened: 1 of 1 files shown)


kaaax0815 (approved):

kaaax0815:
A client could be enabled via the post route too but as its not used in the ui, i think it doesnt matter

kaaax0815:
there is a formatting issue that should be fixed

PR #2625, opened 2026-05-19
feat: add Google OAuth login support

dm-fotc:
## Description

Add Google OAuth login support alongside the existing username/password authentication. When enabled via environment variables, a "Sign in with Google" button appears on the login page. Classic login remains always available regardless of the OAuth configuration.

Key changes:
- New server endpoints for Google OAuth flow (`/api/auth/google`, `/api/auth/google/callback`, `/api/auth/methods`)
- Added `google_id` column to users table with migration
- User service extended with `findOrCreateByGoogle` (auto-links accounts by email)
- CSRF protection via OAuth state parameter stored in session
- Optional domain restriction via `OAUTH_GOOGLE_ALLOWED_DOMAIN`

## Motivation and Context

Users requested the ability to log in via Google accounts for convenience and centralized identity management, while retaining classic username/password login for environments without Google Workspace.

## How has this been tested?

- Configured Google OAuth Client ID/Secret in docker-compose environment
- Tested full OAuth flow: redirect → Google consent → callback → session creation → redirect to dashboard
- Verified domain restriction blocks unauthorized email domains (403)
- Verified existing username/password login still works when OAuth is enabled
- Verified OAuth button is hidden when `OAUTH_GOOGLE_ENABLED` is not set or `false`
- Verified account linking: Google login with email matching existing user links the accounts
- Ran `pnpm lint` and `pnpm format:check` — no issues

image

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,20 +3,18 @@ volumes:
 
 services:
   wg-easy:
-    environment:
+    # environment:
     #  Optional:
-      - PORT=51821
-      - HOST=0.0.0.0
-      - INSECURE=false
+    #  - PORT=51821
+    #  - HOST=0.0.0.0
+    #  - INSECURE=false
     #  Google OAuth (optional):
-      - OAUTH_GOOGLE_ENABLED=true
-      - OAUTH_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
-      - OAUTH_GOOGLE_CLIENT_SECRET=your-client-secret
-      - OAUTH_GOOGLE_ALLOWED_DOMAIN=example.com
+    #  - OAUTH_GOOGLE_ENABLED=true
+    #  - OAUTH_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
+    #  - OAUTH_GOOGLE_CLIENT_SECRET=your-client-secret
+    #  - OAUTH_GOOGLE_ALLOWED_DOMAIN=example.com
 
-    build:
-      dockerfile: ./Dockerfile
-    platform: linux/amd64
+    image: ghcr.io/wg-easy/wg-easy:15
     container_name: wg-easy
     networks:
       wg:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,11 +8,6 @@ services:
     #  - PORT=51821
     #  - HOST=0.0.0.0
     #  - INSECURE=false
-    #  Google OAuth (optional):
-    #  - OAUTH_GOOGLE_ENABLED=true
-    #  - OAUTH_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
-    #  - OAUTH_GOOGLE_CLIENT_SECRET=your-client-secret
-    #  - OAUTH_GOOGLE_ALLOWED_DOMAIN=example.com
 
     image: ghcr.io/wg-easy/wg-easy:15
     container_name: wg-easy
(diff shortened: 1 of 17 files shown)


src/server/database/sqlite.ts
```diff
@@ -71,6 +72,21 @@ async function migrate() {
...
+    DB_DEBUG('Added missing google_id column');
+  } catch {
+    // Column already exists — expected after successful migration
+  }
+  // Ensure all Google OAuth users have ADMIN role
+  await client.execute(
+    "UPDATE users_table SET role = 1 WHERE google_id IS NOT NULL AND role != 1"
+  );
+}
+
```
kaaax0815: the first sql statement is already ensured with the migration.
the second one by the user service

docker-compose.yml
```diff
@@ -3,11 +3,16 @@ volumes:
...
+    # environment:
     #  Optional:
     #  - PORT=51821
     #  - HOST=0.0.0.0
     #  - INSECURE=false
+    #  Google OAuth (optional):
+    #  - OAUTH_GOOGLE_ENABLED=true
+    #  - OAUTH_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
+    #  - OAUTH_GOOGLE_CLIENT_SECRET=your-client-secret
+    #  - OAUTH_GOOGLE_ALLOWED_DOMAIN=example.com
```
kaaax0815: remove them here. in the docs is enough

kaaax0815 (changes requested):

src/server/database/sqlite.ts
dm-fotc: done

kaaax0815:
I did not test it yet, but the code looks good.

But if we want to support oauth, we should make it more provider agnostic, as then users would want providers other than google (e.g. github)

dm-fotc:
> I did not test it yet, but the code looks good.
> 
> But if we want to support oauth, we should make it more provider agnostic, as then users would want providers other than google (e.g. github)

I agree that, long term, it makes sense to keep the OAuth implementation provider-agnostic, especially if we want to support providers like GitHub later. For the first step, though, I’d like to implement sign-in with Google first, and then use that as the basis for extracting a provider abstraction once we add more OAuth providers.

docs/content/advanced/config/optional-config.md
```diff
@@ -11,6 +11,28 @@ You can set these environment variables to configure the container. They are not
...
+3. Add Authorized redirect URI: `https://<your-domain>/api/auth/google/callback`
+4. Copy the Client ID and Client Secret to the environment variables
+
+If a user logs in with Google and their email matches an existing account, the accounts are automatically linked.
+
+///
+
 /// note | IPv6 Caveats
 
 Disabling IPv6 will disable the creation of the default IPv6 firewall rules and won't add a IPv6 address to the interface and clients.
```
kaaax0815, line 38: this should stay with table above.

maybe it even makes sense to move oauth to its own page

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ volumes:
 
 services:
   wg-easy:
-    #environment:
+    # environment:
```
kaaax0815: remove the space

kaaax0815 (changes requested):

kaaax0815:
Some notes:
the table schema should already support multiple providers to make it easier down the road
the oauth completely skips 2fa
pkce is highly recommended nowadays
using a package like `openid-client` should make things easier

These do not have to be handled in this pr. After you fix the changes as requested above I will merge into the dev branch

Thank you very much for the work

docs/content/advanced/config/optional-config.md
kaaax0815: @dm-fotc Please move the oauth block under the ipv6 block. or move it to its own page

kaaax0815:
Thank you very much

Fur future contributions, avoid creating commits on the main/master branch (create a branch). This makes it easier for maintainers to edit PRs

PR #2688, opened 2026-07-10
fix: create firewall rules without comments when xt_comment is unavailable

TowyTowy:
## Description

The per-client firewall feature builds iptables rules with `-m comment --comment "client N: name"`. On hosts whose kernel lacks the `xt_comment` module (e.g. Synology DSM 7.3), iptables parses the argument but rejects the rule at insertion ("Extension comment revision 0 not supported, missing kernel module?"). Because rules are applied during `WireGuard.saveConfig()`, this throws and fails the entire `POST /api/admin/interface` request, making the feature impossible to enable.

This PR detects, once and cached, whether iptables/ip6tables can actually insert a rule using the comment module (the userspace extension may exist while the kernel module does not, so `--help` is not enough — it probes by inserting a throwaway rule into a temporary chain). When comments are unsupported, rules are built without `-m comment`, so filtering keeps working and only the cosmetic client label is dropped. This mirrors the existing graceful degradation when iptables is entirely unavailable.

The rule-building logic is extracted into a pure `buildClientRuleCommands` helper so it can be unit tested.

## Motivation and Context

Fixes #2545.

## How has this been tested?

- New unit tests in `src/test/unit/firewall.spec.ts` covering comment inclusion, the no-comment fallback (regression test asserting no rule uses `-m comment`), IPv6 routing/skip, and the firewallIps → allowedIps → defaultAllowedIps fallback.
- `pnpm test:unit` (33 passed), `pnpm lint`, `pnpm typecheck`, `pnpm format:check` all pass locally.

## Types of changes

## Checklist

## AI disclosure

This change was prepared with AI assistance (Claude). I have reviewed, tested, and understand it, and take responsibility for its correctness.

### DIFF
--- a/src/server/utils/firewall.ts
+++ b/src/server/utils/firewall.ts
@@ -16,6 +16,11 @@ let rebuildQueued = false;
 // Cache iptables availability check result
 let iptablesAvailable: boolean | null = null;
 
+// Cache whether iptables supports the comment module (xt_comment kernel module).
+// Some systems (e.g. Synology DSM) ship iptables without the xt_comment kernel
+// module, which makes `-m comment` fail at rule insertion. See #2545.
+let commentSupported: boolean | null = null;
+
 type ParsedEntry = {
   ip: string;
   port?: number;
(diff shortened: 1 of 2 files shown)


TowyTowy:
Still relevant — the branch applies cleanly on master. Happy to make changes if a maintainer wants a different approach.

PR #2693, opened 2026-07-15
feat(auth): optional trusted-header SSO (reverse-proxy authentication)

culpur:
## Summary

Adds **optional trusted-header SSO** so wg-easy can run behind an authenticating reverse proxy (Authentik, Authelia, oauth2-proxy, …) and trust the identity the proxy has already verified — instead of showing its own login form to users who are already authenticated upstream.

When enabled, the proxy asserts the username via a header, and wg-easy establishes a normal session for that user, provisioning the account on first sight. This is a common request for teams that already run SSO and don't want a second, separate wg-easy credential.

**It is disabled by default. Stock behaviour is completely unchanged unless `TRUSTED_PROXY_ENABLED=true` is set.**

## How it works

Auth still funnels through the existing `getCurrentUser()` chokepoint and the normal stateless session cookie. A small Nitro middleware (`server/middleware/trustedAuth.ts`) runs before routes: when enabled and the request proves it came from the trusted proxy, it reads the identity header, resolves or provisions the user, and calls `useWGSession().update({ userId })`. Everything downstream (RBAC via `definePermissionEventHandler`) then works with zero changes to route handlers.

## Security model

A raw client can forge any header, so the trust boundary is a **shared secret the proxy injects** (`TRUSTED_PROXY_SECRET`), verified with a constant-time compare. The feature **fails closed** if enabled without a secret. An optional source-IP allowlist (`TRUSTED_PROXY_IPS`) is available as defense-in-depth, but it is intentionally optional because container NAT often rewrites the peer to a gateway address, making IP alone non-discriminating — so the secret is the real gate.

Provisioned SSO users have no local password (`password` is now nullable, `auth_provider` distinguishes them) and therefore **cannot** authenticate via the password/Basic path — `isPasswordValid` fails closed on a null hash.

## Changes

- `server/middleware/trustedAuth.ts` (new) — the trust gate + session establishment.
- `server/utils/config.ts` — new `TRUSTED_PROXY_*` env (all off/empty by default).
- `server/database/repositories/user/service.ts` — `getOrCreateExternal()` for password-less federated users (never escalates role/enabled on existing users; refreshes email/name).
- `server/database/repositories/user/schema.ts` + migration `0005` — `password` nullable, add `auth_provider` (existing users backfilled to `local`).
- `server/utils/password.ts` — `isPasswordValid` accepts a nullable hash and fails closed.
- Docs: `advanced/config/optional-config.md` table + security note, `.env.example`, and a commented block in `docker-compose.yml`.

## Config

See the new section in `docs/advanced/config/optional-config.md`. Minimal example:

```env
TRUSTED_PROXY_ENABLED=true
TRUSTED_PROXY_SECRET=<openssl rand -hex 32>   # the proxy must inject this
TRUSTED_PROXY_HEADER=x-authentik-username     # or x-forwarded-user, etc.
TRUSTED_PROXY_DEFAULT_ROLE=client             # or admin
```

## Testing

Built from source and verified end-to-end behind an Authentik forward-auth outpost:
- Valid secret + username header → user auto-provisioned, session established, lands straight in the dashboard.
- Wrong/missing secret → rejected (fails closed), falls back to the normal login form.
- Feature disabled → header ignored entirely, stock behaviour.
- Migration applies cleanly to an existing DB; existing users backfill to `auth_provider=local` and keep working with password login.

Happy to adjust naming, defaults, or split this differently — feedback welcome.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,31 @@
+# wg-easy environment variables
+# Copy to `.env` and adjust. All values below are optional unless noted.
+# Full reference: docs/content/advanced/config/optional-config.md
+
+# --- Web UI ---
+# PORT=51821
+# HOST=0.0.0.0
+# INSECURE=false
+# DISABLE_IPV6=false
+# DISABLE_VERSION_CHECK=false
+
+# --- Trusted-header SSO (Culpur fork) ---
+# Trust an upstream reverse proxy (e.g. an Authentik forward-auth outpost) that
+# has already authenticated the user and asserts their identity via headers.
+# Disabled unless TRUSTED_PROXY_ENABLED=true. Reuses the OAuth account machinery
+# (oauth_provider=trusted-header) and the OAUTH_AUTO_REGISTER gate.
+#
+# SECURITY: the shared secret is the trust boundary — Docker SNAT makes the peer
+# IP non-discriminating, so the secret (not the source IP) proves a request came
+# from the proxy. The feature fails closed if enabled without a secret. Your
+# proxy MUST strip client-supplied identity headers and set its own.
+# TRUSTED_PROXY_ENABLED=true
+# TRUSTED_PROXY_SECRET=change-me-to-a-long-random-value
+# TRUSTED_PROXY_SECRET_HEADER=x-wg-proxy-secret
+# TRUSTED_PROXY_HEADER=x-authentik-username
+# TRUSTED_PROXY_EMAIL_HEADER=x-authentik-email
+# TRUSTED_PROXY_NAME_HEADER=x-authentik-name
+# TRUSTED_PROXY_IPS=
+# TRUSTED_PROXY_DEFAULT_ROLE=client
+# Auto-provision users the proxy authenticates (shared with OAuth):
+# OAUTH_AUTO_REGISTER=true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,18 @@ services:
     #  - PORT=51821
     #  - HOST=0.0.0.0
     #  - INSECURE=false
+    #  Trusted-header SSO (trust an upstream forward-auth proxy, e.g. Authentik).
+    #  Disabled unless TRUSTED_PROXY_ENABLED=true. The shared secret is the trust
+    #  boundary — see docs/advanced/config/optional-config.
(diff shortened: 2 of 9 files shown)


kaaax0815:
`TRUSTED_PROXY_SECRET` is not something I've ever seen. Specifying the Trusted Proxy Ip should be enough

culpur:
Thanks for taking a look, and congrats on landing #2659 — native OAuth/OIDC is great, and it covers the case where wg-easy itself is the OIDC client.

This PR targets a different (complementary) deployment: **forward-auth**, where an authenticating reverse proxy (Authentik/Authelia/oauth2-proxy in `auth_request` / outpost mode) has *already* authenticated the user and asserts the identity via a header — wg-easy trusts that instead of running its own OIDC flow. I'll rebase this on top of your new `oauth_provider` schema and re-scope it to just the forward-auth middleware (dropping the now-redundant nullable-password/column bits you already added).

On `TRUSTED_PROXY_SECRET` vs IP-only: I'd gently push back, from a concrete failure I hit deploying this. With Docker's default userland-proxy port publishing, **every inbound connection is SNAT'd to the bridge gateway IP** (e.g. `172.17.0.1`) before it reaches the container — so `getRequestIP()` returns the gateway for *all* callers, and an IP allowlist can't distinguish the trusted proxy from any other host that can reach the published port. In that (very common) topology, IP-only is not actually a trust boundary. The shared secret is what the proxy injects and only the proxy knows, so it holds even when the peer IP is non-discriminating.

I've kept the IP allowlist as an optional second factor for setups where the peer *is* preserved (host networking, non-NAT), but made the secret the primary gate and fail-closed. Happy to make the secret optional/opt-out if you'd prefer IP-only to remain a supported mode — just let me know which way you'd like it and I'll adjust. Rebase + re-scope incoming.

kaaax0815:
I use many docker containers where i use forward auth with traefik and authelia. the traeffic happens through docker networks and therefore you get the correct source ip of (in this case) the traefik container.

PS: your comments reads like ai, please don't use ai for comments and other stuff [^1]

[^1]: https://github.com/wg-easy/wg-easy/blob/master/contributing.md#ai-contribution-policy

culpur:
Tracking your comments about AI usage for commenting. just ensuring I am being precise.

PR #2701, opened 2026-07-20
fix: skip ip6tables when IPv6 is disabled

### LINKED ISSUE #2698: [Bug]: Firewall rebuild invokes ip6tables when IPv6 is disabled
## Summary

With `DISABLE_IPV6=true`, enabling or rebuilding the per-client firewall still executes an `ip6tables` jump-rule command. On installations where only the documented IPv4 requirement (`iptables`) is present, the rebuild fails and the requested firewall state is not applied.

## Reproduction

1. Use a host/container where `iptables` is available and `ip6tables` is absent.
2. Start wg-easy with `DISABLE_IPV6=true`.
3. Enable the per-client firewall or trigger a configuration save.
4. `firewall.isAvailable(false)` succeeds because it correctly checks only `iptables`, but `initChain()` unconditionally runs the IPv6 jump-rule command and rejects.

I reproduced the exact command path on the test host by removing `/usr/sbin` from `PATH`; it exits 127.

## Environment

- wg-easy commit: `78b283849895b35ad45c6f83fa01228c0f1a6774`
- Ubuntu 24.04.4 LTS, Linux 6.17.0-1018-oracle, ARM64
- Docker host deployment

## Suggested fix

Pass `enableIpv6` into the chain-initialization/removal helpers and skip every `ip6tables` command when it is false. Add a unit test that makes any `ip6tables` invocation fail and verifies an IPv4-only rebuild succeeds.

### Expected behavior

When `DISABLE_IPV6=true`, per-client firewall initialization and rebuilds should execute only IPv4 `iptables` commands and should succeed without `ip6tables` installed.

### Relevant log output

```shell
src/server/utils/firewall.ts:182:
ip6tables -C FORWARD -i <interface> -j WG_CLIENTS 2>/dev/null || ip6tables -I FORWARD 1 -i <interface> -j WG_CLIENTS

Host reproduction:
bash: line 1: ip6tables: command not found
ipv6_disabled_init_exit=127
```

potatosips:
## Description

Threads the existing IPv6-enabled setting through firewall chain initialization, flushing, and removal. IPv4 rules are still maintained normally, while `ip6tables` is never invoked when IPv6 is disabled.

## Motivation and Context

The availability check correctly skipped `ip6tables` in IPv4-only mode, but later chain-management methods invoked it unconditionally. On an IPv4-only image without that executable, firewall rebuilds failed after the successful availability check.

Closes #2698.

## How has this been tested?

- Added unit tests covering initialization, flushing, and removal in IPv4-only mode.
- Full unit suite: 31 passed.
- ESLint and Nuxt typecheck passed locally and on Ubuntu 24.04 ARM64 in a Node 24 container.
- Reproduced the pre-fix command failure (`exit 127`) on the test host before applying the change.

Not applicable.

## Types of changes

## Checklist:

AI assistance was used during review, implementation, and drafting. The resulting code and tests were reviewed and verified against the reported reproduction.

### DIFF
--- a/src/server/utils/firewall.ts
+++ b/src/server/utils/firewall.ts
@@ -164,32 +164,38 @@ export const firewall = {
   /**
    * Initialize the custom chain if it doesn't exist
    */
-  async initChain(interfaceName: string): Promise<void> {
+  async initChain(interfaceName: string, enableIpv6: boolean): Promise<void> {
     FW_DEBUG(
       `Initializing firewall chain ${CHAIN_NAME} for interface ${interfaceName}`
     );
 
     // Create chain if not exists (iptables returns error if exists, so we ignore)
     await exec(`iptables -N ${CHAIN_NAME} 2>/dev/null || true`);
-    await exec(`ip6tables -N ${CHAIN_NAME} 2>/dev/null || true`);
+    if (enableIpv6) {
+      await exec(`ip6tables -N ${CHAIN_NAME} 2>/dev/null || true`);
+    }
 
     // Ensure chain is referenced from FORWARD (if not already)
     // Insert at position 1 to process before generic ACCEPT rules
     await exec(
       `iptables -C FORWARD -i ${interfaceName} -j ${CHAIN_NAME} 2>/dev/null || iptables -I FORWARD 1 -i ${interfaceName} -j ${CHAIN_NAME}`
     );
-    await exec(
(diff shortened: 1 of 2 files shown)


kaaax0815 (approved):

kaaax0815:
Thank you for the PRs

PR #2707, opened 2026-07-21
perf: avoid repeated client scans during CIDR changes

### LINKED ISSUE #2706: [Performance]: CIDR changes repeatedly scan all clients
## Summary

Changing an interface CIDR performs a full `clients` query inside the loop over every client, then calls `nextIP`, which scans the address pool and performs a linear client lookup for each candidate. The result scales superlinearly and can stall a CIDR update on installations with many clients.

## Reproduction

1. Start from upstream commit `f731135564ca92944a1106d280bbbd1fb7a2cef8`.
2. Configure a WireGuard interface with a sufficiently large client list.
3. Change either CIDR through the interface CIDR endpoint.
4. Observe the repeated `tx.query.client.findMany()` in `InterfaceService.updateCidr`, nested under the loop over clients, and the repeated linear `clients.find()` calls in `nextIP`.

On Ubuntu 24.04 ARM64, a representative execution of the same update loop and address-allocation logic took 764.02 ms for 800 clients and 5,037.25 ms for 1,600 clients. The repeated database reads make the real operation slower still.

## Expected behavior

CIDR changes should scale close to linearly with the number of clients and should avoid re-reading the complete table for each client.

## Actual behavior

The implementation performs one initial full query plus one full query per client, while address selection repeatedly scans the client list.

## Suggested fix

Load clients once per transaction and maintain in-memory sets of allocated IPv4/IPv6 addresses as records are updated. Use set membership during allocation instead of linear searches.

## Environment

- Commit: `f731135564ca92944a1106d280bbbd1fb7a2cef8`
- Test host: Ubuntu 24.04.4 LTS, ARM64

potatosips:
## Description

Avoids repeated full-client reads and linear address lookups while reassigning client addresses after a CIDR change.

## Motivation and Context

`InterfaceService.updateCidr` queried every client inside its loop and passed the complete list to an allocator that performed another linear search for every candidate address. This made CIDR changes grow superlinearly with the number of clients.

Closes #2706.

## How has this been tested?

- Added unit coverage for set-backed address allocation, address exhaustion, and replacing addresses while retaining allocation state.
- Full unit suite: 37 passed.
- ESLint, Nuxt typecheck, and formatting checks passed.
- Representative VPS benchmark of the original logic: 800 clients took 764.02 ms; 1,600 clients took 5,037.25 ms.

Not applicable.

## Types of changes

## Checklist:


src/server/database/repositories/interface/service.ts
```diff
@@ -93,21 +93,32 @@ export class InterfaceService {
...
-        // TODO: optimize
-        const clients = await tx.query.client.findMany().execute();
-
         // only calculate ip if cidr has changed
 
         let nextIpv4 = client.ipv4Address;
-        if (data.ipv4Cidr !== oldCidr.ipv4Cidr) {
-          nextIpv4 = nextIP(4, parseCidr(data.ipv4Cidr), clients);
+        if (ipv4Cidr) {
+          nextIpv4 = nextIPFromUsedAddresses(4, ipv4Cidr, ipv4Addresses);
```
kaaax0815: keep old check and move parse into function param. should reduce changes

kaaax0815 (changes requested):

src/server/database/repositories/interface/service.ts
potatosips: Addressed in 0aeaa5e: restored the original CIDR-change guards and parse the CIDR directly at the helper call site. Thank you.

kaaax0815 (commented):
please optimize `src/server/api/setup/migrate.post.ts` too

potatosips:
I missed the setup migration path. I’ve prepared the same optimization there: it now loads the existing clients once, tracks assigned IPv6 addresses in a set, and updates that set as each client is migrated. I’ll include it in the next update.

potatosips:
Included in 1b03721. `migrate.post.ts` now reads existing clients once and tracks IPv6 allocations in a set during import. Thanks for pointing it out.

Issue #2682, opened 2026-06-29
[Bug]: cli dependencies not correctly bundled


yetty:
## Description

The documented wg-easy CLI command fails immediately because the packaged image imports `citty` from `/app/server/cli.mjs`, but the package is not present in the image.

This breaks the official password reset procedure documented at:
https://wg-easy.github.io/wg-easy/latest/guides/cli/#reset-password

## Reproduction

Using the current v15 image:

```sh
docker run --rm ghcr.io/wg-easy/wg-easy:15 cli --help
```

Also reproducible with edge:

```sh
docker pull ghcr.io/wg-easy/wg-easy:edge
docker run --rm ghcr.io/wg-easy/wg-easy:edge cli --help
```

The same error occurs for the documented password reset command:

```sh
docker compose exec -it wg-easy cli db:admin:reset
```

## Actual behavior

```text
node:internal/modules/package_json_reader:301
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'citty' imported from /app/server/cli.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
    at packageResolve (node:internal/modules/esm/resolve:764:81)
    at moduleResolve (node:internal/modules/esm/resolve:855:18)
    at defaultResolve (node:internal/modules/esm/resolve:988:11)
    ...
```

For edge, the same failure occurs with Node v24.17.0.

## Expected behavior

The CLI should start and show help, and the documented reset command should prompt for a new admin password:

```sh
docker compose exec -it wg-easy cli db:admin:reset
```

## Environment

- Image: `ghcr.io/wg-easy/wg-easy:15`
- wg-easy version reported by container logs: `v15.3.0`
- Stable image digest after pull: `sha256:93bbd593e07bab98d02807a28770ac87ab6c48818e319e68c1f66561feb99876`
- Stable image created: `2026-05-18T10:21:22.51442822Z`
- Also tested: `ghcr.io/wg-easy/wg-easy:edge`

## Additional evidence

Inside the stable image, `/app/server/cli.mjs` imports `citty`, but `/app/server/node_modules/citty` is missing.

This appears to be a packaging/dependency issue affecting the CLI entrypoint.

ondrejnedoma:
[...]

PR #2713, opened 2026-07-25
fix(cli): bundle JS dependencies into cli.mjs

mikhailde:
## Description

This PR resolves an issue where running CLI commands (such as `cli db:admin:reset`) inside the production Docker image fails with an `ERR_MODULE_NOT_FOUND` error for dependencies like `citty` and `consola`.

While the previous PR (#2683) introduced dependency tracing via Nitro's `traceInclude`, multi-stage Docker builds aggressively prune `node_modules` in the final image, causing standalone scripts like `cli.mjs` to lose access to external packages.

To make the CLI truly resilient and self-contained, this change switches the strategy from **tracing** to **bundling**:
- **`src/cli/build.js`**: Enabled `bundle: true` in `esbuild` for pure JS dependencies while keeping native binaries (`argon2`, `@libsql/client`) external. Removed the `--trace` workaround.
- **`src/nuxt.config.ts`**: Cleaned up obsolete `nitro.externals.traceInclude` logic.
- **`src/package.json`**: Streamlined the build pipeline by removing the intermediate `cli:trace` step.

## Motivation and Context

- Fixes `ERR_MODULE_NOT_FOUND` during CLI executions in production containers.
- Addresses the root cause behind #2682.
- Makes the CLI environment-agnostic, ensuring it no longer depends on fragile `node_modules` paths inside the Docker container.

## How has this been tested?

1. **Docker Build:** Successfully built a local image (`docker build -t wg-easy-local .`).
2. **CLI Execution:** Tested `docker run --rm wg-easy-local cli --help` and `docker run --rm wg-easy-local cli db:admin:reset --help`. Both executed cleanly and displayed expected CLI prompts without missing module errors.
3. **Web UI Regression:** Verified that the main Nuxt server compiles without issues and starts as expected (`Listening on http://0.0.0.0:51821`).

## Types of changes

- [x] Refactoring (change which does not alter existing functionality)
- [ ] Documentation (changes to documentation only)

## Checklist:

## AI Disclosure:

Google Gemini was used to assist in analyzing the build pipeline, diagnosing the multi-stage Docker interaction with Nitro, and drafting this PR. All code changes were thoroughly reviewed, manually compiled, and verified in a local Docker build prior to submission.

### DIFF
--- a/src/cli/build.js
+++ b/src/cli/build.js
@@ -1,58 +1,18 @@
 // @ts-check
 
-import { writeFile, mkdir } from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
-
 import esbuild from 'esbuild';
 
-const isTrace =
-  process.argv.slice(2).findIndex((arg) => arg === '--trace') !== -1;
+// Native modules that cannot be bundled and are handled separately in Docker
+const external = ['argon2', '@libsql/client'];
 
-/**
- * Writes files with needed imports to `../node_modules/.cache/wg-easy/trace.mjs`
- *
- * @param {import('esbuild').Metafile} metafile
- */
-async function writeTrace(metafile) {
-  const paths = [
-    ...new Set(
-      Object.values(metafile.outputs).flatMap((v) =>
-        v.imports.map((v) => v.path)
-      )
-    ),
-  ];
-  const imports = paths.map((v) => `import '${v}';`).join('\n');
-
-  const folderUrl = new URL('../node_modules/.cache/wg-easy/', import.meta.url);
-  const folder = fileURLToPath(folderUrl);
-  const filePath = fileURLToPath(new URL('./trace.mjs', folderUrl));
-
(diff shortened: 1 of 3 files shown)


kaaax0815:
the cli works and i think tracing is the correct solution

mikhailde:
Thanks for taking a look @kaaax0815!

Just to follow up after investigating this further: I was testing against the current stable release image (`v15.3.0`), where the CLI fails with `ERR_MODULE_NOT_FOUND` because #2683 hasn't been published in a stable release tag yet.

I just tested against the `edge` image (`v15.4.0-beta.1`) and can confirm the CLI works as expected there.

Since the CLI is currently broken for everyone on the `v15.3.0` stable tag, it might be worth releasing a patch (`v15.3.1`) or tagging `v15.4.0` when ready, as users on the stable release can't use documented CLI commands until then.

Thanks for maintaining wg-easy!

PR #2721, opened 2026-08-02
feat: Add ability to set Table parameter in wg0.conf

### LINKED ISSUE #2708: [Feat]: Support Table option in WireGuard interface configuration
## Summary

It would be great if `wg-easy` exposed the WireGuard `Table` option (used by `wg-quick`) as an optional interface setting.

Currently there is no way to configure it through the web UI or environment variables.

## Why this is useful

The `Table` option is required for more advanced routing scenarios, for example:

- Policy routing
- Custom routing tables
- Multi-hop WireGuard setups
- Split tunneling with manual routing

Without this option, `wg-quick` always uses its default behavior, making these configurations difficult or impossible to implement cleanly.

## Current workaround

At the moment, the only workaround I found is to override the internal `wg.conf` template generated by `nitro.mjs`.

This requires:

1. Copying the template from the container.
2. Adding a single line:

   ```ini
   Table = 123
   ```

3. Bind-mounting the modified template back into the container.

While this works, it has a major downside: every image update may change the template, requiring users to manually compare and update their custom copy.

It also means overriding an internal project file just to use a standard `wg-quick` configuration option.

### Suggested Solution

Expose the `Table` option as an optional setting, for example:

- in the web UI (Interface settings), or
- through an environment variable such as `WG_TABLE`.

If not specified, the current behavior (`auto`) would remain unchanged, making this feature fully backward compatible.

This small addition would eliminate the need for maintaining custom templates while enabling advanced routing configurations already supported by `wg-quick`.

Thanks for considering it!

### Alternatives

_No response_

### Additional Context

_No response_

vlad-iakovlev:
## Description

- Adds UI and DB fields to control `Table` parameter in `wg0.conf`

## Motivation and Context

I use configuration with one of the clients as exit node. For it to work I need to disable automatic route management which can be done by adding `Table=off` to the `wg0.conf`.

## How has this been tested?

1. Deployed to existing server, verified that the parameter sets to `auto` (in UI, DB and `wg0.conf`) and VPN works
2. Deployed to another existing server, changed parameter to `off`, verified in UI, DB and `wg0.conf` and checked that VPN works with expected exit node
3. Launched new server, verified that the parameter sets to `auto` (in UI, DB and `wg0.conf`) and VPN works

## Screenshots:

image

## Types of changes

- [ ] Refactoring (change which does not alter existing functionality)
- [ ] Documentation (changes to documentation only)

## Checklist:

## AI Disclosure:

I've used Cursor with Composer model to explore the project and generate the solution. After that I reviewed and, where needed, corrected all generated code.


kaaax0815:
Thank you

PR #2779, opened 2026-08-26
Feature/add description field and tag

codyq:
## Description

Adds two related features to client management:

1. **Client description** — clients can now have an optional free-text `description` alongside their `name`, editable from the client create/edit pages and shown on the home page client list.

2. **Tags** — a new many-to-many tagging system for clients:
   - New `tags_table` + `client_tags_table` join table (composite PK, cascading FKs) — the first many-to-many relation in the schema.
   - New `/api/tag` endpoints: `GET`/`POST /api/tag`, `POST`/`DELETE /api/tag/[tagId]`. Viewing tags is open to any authenticated user; creating/editing/deleting is admin-only.
   - `GET /api/client` now accepts repeatable `tagId`/`tagName` query params (OR-matched, can be combined) to filter clients by tag.
   - Client GET/POST responses now include/accept `tags`/`tagIds`; updates use replace-all semantics (submit the full tag set, server diffs it).
   - New Admin Panel → Tags section: list, create, edit, delete tags, and per-tag a checklist to assign/unassign clients.
   - New tag picker on the client create/edit pages: search-and-select existing tags, plus inline tag creation for admins.

**Technical decisions worth flagging for review:**
- The dev environment used for this work had no `node_modules`/`pnpm` available at the time the DB migration was written, so `0009_wandering_paladin.sql` and its Drizzle snapshot were hand-authored (mirroring the format of `drizzle-kit generate`'s output) rather than generated. I later got `pnpm` working and ran the full check suite against it successfully, but a `pnpm db:generate` diff against this migration would be a good sanity check in review.
- Adding this feature's new routes and richer `/api/client` response types pushed Nuxt/Nitro's auto-generated route-matching type past TypeScript's internal recursion limit (`TS2321: Excessive stack depth`), which broke type inference at 20 `$fetch()` call sites **app-wide**, most in files this feature doesn't otherwise touch (`me.vue`, `setup/*.vue`, `admin/*.vue`, `login/*.vue`, `stores/auth.ts`, etc.). I fixed all 20 by giving each call an explicit response-type generic (`$fetch<T>(...)`), which is the standard way to sidestep this class of Nitro/TypeScript scaling issue. This is a real but mechanical fix — see the "unrelated modifications" note below.

## Motivation and Context

Managing more than a handful of WireGuard clients gets hard to navigate with just a name. Description gives admins a place for free-text context (who/what a client is), and tags let clients be grouped and filtered — both in the UI and via the API (e.g. scripting against `?tagName=` to act on a subset of clients) — which name-based search alone doesn't support well.

Not tied to an existing issue.

## How has this been tested?

- **Automated checks** (all passing locally): `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm format:check:docs`, `pnpm build`, `pnpm test:unit` (42/42).
- **Manual end-to-end testing** against `docker-compose.dev.yml`: logged in via `/api/auth/password`, then exercised the API directly (`curl`) — tag create/list/rename/delete, duplicate-tag-name rejection, client create/update with `tagIds` (including replace-all semantics), tag filtering via `tagId`/`tagName` (single value, repeated/OR, and combined), cascade delete (deleting a tag removes it from all clients via FK cascade, verified via the API), and permission enforcement (401 without a session).
- **Manual UI testing** in a browser against the same Docker instance — client create/edit tag picker (search, assign, inline create), Admin Panel → Tags (list, create, edit, delete, per-tag client checklist), home page tag display — with a few rounds of layout/styling fixes based on what didn't look right in practice (grid alignment, popover styling, chip styling).

## Screenshots

_(Not attached here — I don't have the browser screenshots as files; happy to add the client tag picker / Admin Panel → Tags screenshots if useful for review.)_

## Types of changes

## Checklist

- [ ] I have reviewed my own changes.
- [ ] My changes do not include unrelated modifications.

## AI Disclosure

Implemented with substantial assistance from Claude (Claude Code / Sonnet 5, Anthropic): schema/migration design, API endpoints, permission logic, Vue components, and diagnosis + fix of the TS2321 typecheck regression described above were all AI-drafted, directed iteratively by me (requirements, permission model decisions, UI feedback from my own screenshots, scope calls on docs/tests). I've read through the implementation and tested it manually as described above; I have not yet added automated tests for the new tag logic.


kaaax0815 (changes requested):
only create one migration. create migration with `pnpm db:generate`.
remove unrelated changes.

does it make sense to make global tags? or would user specific tags make more sense for the future user system?

codyq:
I've squashed the two migrations into one and regenerated with pnpm db:generate, and reverted the unrelated changes.

In terms of tags, I think global makes more sense here. A VPN setup is typically per-organisation, so everyone should be working off the same set of tags rather than each user maintaining their own. Per-user tags also mean extra administrative overhead — every user has to set up and maintain their own tag list — and it's error-prone, since nothing keeps them consistent. For example, a device actually located in the US could end up tagged "UK" by John and "Singapore" by Mary, with nothing to catch the discrepancy. With global tags, the sysadmin defines and assigns "US" once, that assignment gets checked for correctness, and everyone views and filters against that same verified source of truth.

This is already how it's implemented — tags_table has no userId/owner column, and access is role-based: ADMIN can create/update/delete tags, CLIENT can only view and filter by them. So no code changes needed there, just confirming the design is correct as-is.

codyq:
Here are the screenshots of those two features

image

image

image

image

PR #2822, opened 2026-09-16
fix: omit private keys from client detail API

SebTardif:
## Description

`GET /api/client/:id` no longer includes `privateKey` or `preSharedKey`.

The handler used to spread the full database row. List endpoints already omit those columns (`getAllPublic` / `getAllForUser`). This change uses the same omit for the detail GET. Config download and QR stay on `/api/client/:id/configuration` and `/qrcode.svg`.

## Motivation and Context

[#2015](https://github.com/wg-easy/wg-easy/pull/2015) stopped returning keys from `GET /api/client` because that list is polled often. The detail GET still returned both secrets. The client edit page never reads them; save goes through `ClientUpdateSchema`, which does not include those fields.

Leaving keys on the metadata GET puts them in browser JSON, proxies, and logs even when the user only opened the edit form.

The current return shape was introduced when endpoint was added in [#2058](https://github.com/wg-easy/wg-easy/pull/2058) (2025-07-25).

## How has this been tested?

```
cd src && pnpm test:unit
```

60 tests passed. Red-green on `test/unit/clientPublic.spec.ts`: without the omit the assertion failed because `privateKey` and `preSharedKey` were still present; with the omit it passed.

Also ran `eslint` and `prettier --check` on the changed files, and `pnpm typecheck`.

## Types of changes

## Checklist

- [x] I have added or updated tests where appropriate.


kaaax0815:
please create function `getPublic` in client service to align with existing code

SebTardif:
@kaaax0815

> please create function `getPublic` in client service to align with existing code

Done. `ClientService.getPublic` excludes `privateKey` and `preSharedKey` in the query, the same way `getAllPublic` does. `GET /api/client/:id` calls that method. Config download still uses `get`, which keeps the secrets.

